PR #29, opened 2021-12-28
feat: add multicall handler

nchamo:
We are now making the first step towards adding the multicall handler. The idea is to support multiple calls in one tx, like:
- Give the companion permissions and withdraw as eth
- Execute withdraws from swapped and unswapped balance in one tx, without terminating

In order to support these use cases, we need to add calls these "proxy calls" in the companion, so we can then use them in the multicall


alejoamiras (approved):
Nice ! One thing, that maybe it's a bit too late to talk about. I feel that having all the functions with `Proxy` might add size, and overhead that we might avoid using a proxy with a contract with delegatecall.
image

Other than that sugg. this is awesome ! And it will allow a lot of cool stuff !

alejoamiras:
Catched up via DM - current design is OK. Disregard my prev. comment. 🚀

PR #30, opened 2021-12-28
feat: add permission permit proxy

nchamo:
We are now adding the last proxy, the ability to add permission permits


alejoamiras (approved):
looks gucci !

PR #31, opened 2021-12-29
feat: implement multicall

nchamo:
We now implemented `Multicall` and added some tests that use it

### DIFF
--- a/deploy/000_register_hub.ts
+++ b/deploy/000_register_hub.ts
@@ -9,6 +9,7 @@ import {
 } from '@mean-finance/dca-v2-core/artifacts/contracts/oracles/ChainlinkOracle.sol/ChainlinkOracle.json';
 import { HardhatRuntimeEnvironment } from 'hardhat/types';
 import { DeployFunction } from 'hardhat-deploy/types';
+import { BigNumber } from 'ethers/lib/ethers';
 
 const deployFunction: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // TODO: Use address from npm package when exposed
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -56,7 +56,7 @@
     "postpublish": "pinst --enable"
   },
   "dependencies": {
-    "@mean-finance/dca-v2-core": "^1.0.0-rc.8"
+    "@mean-finance/dca-v2-core": "^1.0.0-rc.9"
```
nchamo, line 59: Needed to update to the latest version that doesn't have the bug

alejoamiras (approved):
This is awesome ! Great work ... One Q: Would you endulge me with one "attack test" ? Trying to permit / modify / withdraw something thats not owned by the user ? 
Better to be safe than encounter some weird loophole ? 🥺

deploy/000_register_hub.ts
```diff
@@ -28,7 +29,7 @@ const deployFunction: DeployFunction = async function (hre: HardhatRuntimeEnviro
     const chainlinkDeployment = await hre.deployments.deploy('Oracle', {
       from: deployer,
       contract: { abi: CHAINLINK_ABI, bytecode: CHAINLINK_BYTECODE },
-      args: [WETH, CHAINLINK_REGISTRY, governor],
+      args: [WETH, CHAINLINK_REGISTRY, BigNumber.from(2).pow(32).sub(1), governor],
```
alejoamiras, line 32: also, this is due to a dca-v2-core change, right ?

nchamo: Indeed

PR #32, opened 2021-12-29
feat: implement deposit proxy

nchamo:
Initially, I though there was no need for a deposit proxy. But then I realized there was a good use case for it.

What if someone wanted to withdraw funds from their position and create a new position with that balance? So I created the deposit proxy, so it can all be done in one tx

### DIFF
--- a/contracts/DCAHubCompanion/DCAHubCompanionMulticallHandler.sol
+++ b/contracts/DCAHubCompanion/DCAHubCompanionMulticallHandler.sol
@@ -19,6 +19,23 @@ abstract contract DCAHubCompanionMulticallHandler is Multicall, DCAHubCompanionP
     permissionManager.permissionPermit(_permissions, _tokenId, _deadline, _v, _r, _s);
   }
 
+  function depositProxy(
+    address _from,
+    address _to,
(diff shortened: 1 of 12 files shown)


contracts/DCAHubCompanion/DCAHubCompanionMulticallHandler.sol
```diff
@@ -19,6 +19,23 @@ abstract contract DCAHubCompanionMulticallHandler is Multicall, DCAHubCompanionP
...
+    address _from,
+    address _to,
+    uint256 _amount,
+    uint32 _amountOfSwaps,
+    uint32 _swapInterval,
+    address _owner,
+    IDCAPermissionManager.PermissionSet[] calldata _permissions,
+    bool _transferFromCaller
+  ) external returns (uint256 _positionId) {
+    if (_transferFromCaller) {
```
nchamo, line 32: This is here because if we are using balance that was withdrawn on the same multicall, we already have the balance in the contract 🔥

alejoamiras (approved):
Nice ! Actually someone was suggesting that use case in our b-testers channel !

PR #39, opened 2022-01-12
refactor: keep3r job no longer needs swapper to be set in deployment

alejoamiras:
In order to be able to deploy the keep3r job and start acquiring keep3r credits through `V2` we need to deploy the keep3r job alone. This PR solves that.

### DIFF
--- a/contracts/DCAKeep3rJob/DCAKeep3rJob.sol
+++ b/contracts/DCAKeep3rJob/DCAKeep3rJob.sol
@@ -13,13 +13,8 @@ contract DCAKeep3rJob is Governable, IDCAKeep3rJob {
   address public swapper;
   mapping(address => bool) public canAddressSignWork;
 
-  constructor(
-    address _swapper,
-    IKeep3rJobs _keep3r,
-    address _governor
-  ) Governable(_governor) {
-    if (address(_swapper) == address(0) || address(_keep3r) == address(0)) revert ZeroAddress();
-    swapper = _swapper;
+  constructor(IKeep3rJobs _keep3r, address _governor) Governable(_governor) {
+    if (address(_keep3r) == address(0)) revert ZeroAddress();
     keep3r = _keep3r;
   }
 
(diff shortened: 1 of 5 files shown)


contracts/DCAKeep3rJob/DCAKeep3rJob.sol
```diff
@@ -13,13 +13,8 @@ contract DCAKeep3rJob is Governable, IDCAKeep3rJob {
...
 
-  constructor(
-    address _swapper,
-    IKeep3rJobs _keep3r,
-    address _governor
-  ) Governable(_governor) {
-    if (address(_swapper) == address(0) || address(_keep3r) == address(0)) revert ZeroAddress();
-    swapper = _swapper;
+  constructor(IKeep3rJobs _keep3r, address _governor) Governable(_governor) {
+    if (address(_keep3r) == address(0)) revert ZeroAddress();
```
nchamo, line 17: I would keep the parameter, but allow the zero address. So that it can be set when deployed, but it's just not necessary. Otherwise you'll be forzing an extra tx always

deploy/003_keep3r_job.ts
```diff
@@ -17,15 +18,15 @@ const deployFunction: DeployFunction = async function (hre: HardhatRuntimeEnviro
       return;
   }
 
-  const companion = await hre.deployments.get('DCAHubCompanion');
+  const companion = await hre.deployments.getOrNull('DCAHubCompanion');
```
nchamo, line 21: I have a question about how this would work. If we do this, when we deploy the companion to mainnet, will the keep3r job automatically be re-deployed?

test/integration/DCAKeep3rJob/keep3r-job.spec.ts
```diff
@@ -59,6 +59,8 @@ contract('DCAKeep3rJob', () => {
     const k3prWhale = await wallet.impersonate(K3PR_WHALE_ADDRESS);
     await ethers.provider.send('hardhat_setBalance', [K3PR_WHALE_ADDRESS, '0xffffffffffffffff']);
 
+    // Set companion as swapper
+    // await DCAKeep3rJob.connect(governor).setSwapper(DCAHubCompanion.address);
```
nchamo, line 63:
```suggestion
```

nchamo (approved):
Great work!

PR #43, opened 2022-01-20
feat: add misc

nchamo:
We are making two changes here:
1. We are adding `misc` to all functions that end up executing `deposit`
2. We are no longer supporting special deposits on the companion where the protocol token is `to`

### DIFF
--- a/contracts/interfaces/IDCAHubCompanion.sol
+++ b/contracts/interfaces/IDCAHubCompanion.sol
@@ -135,20 +135,6 @@ interface IDCAHubCompanionSwapHandler is IDCAHubSwapCallee {
 }
 
 interface IDCAHubCompanionWTokenPositionHandler {
-  /// @notice Emitted when a deposit is made by converting one of the user's tokens for another asset
-  /// @param positionId The id of the position that was created
-  /// @param originalTokenFrom The original "from" token
-  /// @param convertedTokenFrom The "from" token that was actually deposited on the hub
-  /// @param originalTokenTo The original "to" token
-  /// @param convertedTokenTo The "to" token that was actually part of the position
-  event ConvertedDeposit(
-    uint256 positionId,
-    address originalTokenFrom,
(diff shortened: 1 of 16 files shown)


test/unit/DCAHubCompanion/dca-hub-companion-multi-call-handler.spec.ts
```diff
@@ -104,30 +104,69 @@ contract('DCAHubCompanionMulticallHandler', () => {
     const AMOUNT_OF_SWAPS = 40;
     const SWAP_INTERVAL = 10000;
     const OWNER = '0x0000000000000000000000000000000000000003';
+    const MISC = ethers.utils.randomBytes(10);
```
alejoamiras, line 107: Beware, this random in tests might affect in a small way gas coverage (not really too important)

contracts/interfaces/IDCAHubCompanion.sol
```diff
@@ -135,20 +135,6 @@ interface IDCAHubCompanionSwapHandler is IDCAHubSwapCallee {
...
-  /// @param convertedTokenFrom The "from" token that was actually deposited on the hub
-  /// @param originalTokenTo The original "to" token
-  /// @param convertedTokenTo The "to" token that was actually part of the position
-  event ConvertedDeposit(
-    uint256 positionId,
-    address originalTokenFrom,
-    address convertedTokenFrom,
-    address originalTokenTo,
-    address convertedTokenTo
-  );
```
alejoamiras, line 150: Bye bye !

alejoamiras (approved):

test/unit/DCAHubCompanion/dca-hub-companion-multi-call-handler.spec.ts
nchamo: Good point!

PR #46, opened 2022-01-27
feat: supporting payable multicalls

nchamo:
Before this change, we couldn't execute payable functions in Multicall. This was a problem, since we couldn't execute `increasePositionUsingProtocolToken`

Now, our custom Multicall supports both payable and non-payable functions


alejoamiras (approved):
Nice catch dude ! Let's keep bludgeoning it.

PR #48, opened 2022-02-02
feat: add keep3r setter

0xged:
I was wondering what would be the gas diff if we didn't have it as immutable.


nchamo (approved):
Great work @0xged !

PR #52, opened 2022-02-24
test: use uniswap oracle for swap tests

nchamo:
We are now making two changes that should make swap tests more reliable:
- We are using the Uniswap oracle as the source of truth. Before, when we used Chainlink, it could happen that the price would be different from the current state of the market
- We are passing `takerAddress` to 0x, which makes it more reliable


alejoamiras (approved):
Nice idea !

PR #58, opened 2022-03-09
test: mai liquidity for integration

0xged:
We are testing liquidity on Polygon via 0x - to prove we will be able to execute trades on polygon for it.


nchamo (approved):
Great that this works!!!

PR #65, opened 2022-04-21
feat: adds tokensProxy when swapping with dex

0xged:


0xged:
@nchamo just check changes on DCAHubCompanion. Most of changed files come from PR #67 that fixed integrations tests.

contracts/interfaces/IDCAHubCompanion.sol
```diff
@@ -105,6 +105,7 @@ interface IDCAHubCompanionSwapHandler is IDCAHubSwapCallee {
 
   /// @notice Executes a swap with the given DEX, and sends all unspent tokens to the given recipient
   /// @param _dex The DEX that will be used in the swap
+  /// @param _tokensProxy The spender of the tokens (we will not always need to approve the dex)
```
nchamo, line 108: I think this is a little more clear
```suggestion
  /// @param _tokensProxy The spender of the tokens (could be different from the dex)
```

contracts/interfaces/IDCAHubCompanion.sol
```diff
@@ -126,6 +128,7 @@ interface IDCAHubCompanionSwapHandler is IDCAHubSwapCallee {
   /// @notice Executes a swap with the given DEX and sends all `reward` unspent tokens to the given recipient.
   /// All positive slippage for tokens that need to be returned to the hub is also sent to the hub
   /// @param _dex The DEX that will be used in the swap
+  /// @param _tokensProxy The spender of the tokens (we will not always need to approve the dex)
```
nchamo, line 131: I think this is a little more clear
```suggestion
  /// @param _tokensProxy The spender of the tokens (could be different from the dex)
```

nchamo (approved):
Looks good!

PR #67, opened 2022-04-25
fix: integration tests deployments

0xged:


deploy/002_beta_migrator.ts
```diff
@@ -29,6 +29,5 @@ const deployFunction: DeployFunction = async function (hre: HardhatRuntimeEnviro
   });
 };
 
-deployFunction.dependencies = ['DCAHub'];
```
0xged, line 32: This didn't make any sense.

hardhat.config.ts
```diff
@@ -118,6 +118,16 @@ const config: HardhatUserConfig = {
...
+    contracts: [
+      {
+        artifacts: 'node_modules/@mean-finance/dca-v2-core/artifacts',
+        deploy: 'node_modules/@mean-finance/dca-v2-core/deploy',
+      },
+      {
+        artifacts: 'node_modules/@mean-finance/chainlink-registry/artifacts',
+        deploy: 'node_modules/@mean-finance/chainlink-registry/deploy',
+      },
+    ],
```
0xged, line 130: Includes deployment scripts from other repositories into `hardhat-deploy`

0xged (commented):
We've changed all the DCAHubCompanions test that we could to use ethereum mainnet, just to leverage the dca-v2-core deployment scripts. Nothing really changes in terms of what its testing.

PR #69, opened 2022-04-29
refactor: remove swap handler and top optimizer

0xged:
Although it has a lot of changed files, it doesn't change really much:
- Moved DCAHubCompanionSwapHandler to its own contract (DCAHubSwapper)
- Fixed tests
- Removed DCAHubCompanionSwapHandler

This allowed us to:
- Take the optimizier up a notch (although not big optimizations ~1%)
- Will allow us to iterate without having users to sign permits to new companion if we need to iterate the swapper


contracts/DCAHubSwapper/DCAHubSwapperParameters.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.7 <0.9.0;
+
+import '../interfaces/IDCAHubSwapper.sol';
+import '../utils/Governable.sol';
+
+abstract contract DCAHubSwapperParameters is Governable, IDCAHubSwapperParameters {
```
nchamo, line 7: Personal and ignorable opinion: I think it would make more sense to move this to the SwapHandler. This extra contract is not needed, and it doesn't add much.

nchamo (approved):
Great work!

PR #71, opened 2022-04-29
refactor: generalize migrator

nchamo:
We are now generalizing the migrator. Previously, it worked only from beta to v2. Now, it will take the addresses of the source and target hub, and migrate from one another.

It will allow us to migrate from beta to new v2, and from vuln v2 to new v2.


test/integration/V2Migration/position-migrator.spec.ts
```diff
@@ -80,41 +79,55 @@ contract.skip('BetaMigrator', () => {
...
+    title: 'migrating from beta version',
+    sourceHub: () => betaDCAHub,
+    targetHub: () => DCAHub,
   });
 
+  migratePositionTest({
+    title: 'migrating from vulnerable version',
+    sourceHub: () => vulnDCAHub,
+    targetHub: () => DCAHub,
+  });
```
0xged, line 92: Ha! Nice!

0xged (approved):
Looks great! Love the combination of BetaMigrator and the one we needed now!

PR #77, opened 2022-05-21
feat: add target tokens distribution

nchamo:
We are now making the tokens distribution configurable. The contract owner (and other allowed users) will be able to set it

### DIFF
--- a/contracts/DCAFeeManager/DCAFeeManager.sol
+++ b/contracts/DCAFeeManager/DCAFeeManager.sol
@@ -1,61 +1,201 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity >=0.8.7 <0.9.0;
 
+import '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+import '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
+import '@openzeppelin/contracts/utils/Multicall.sol';
 import '../interfaces/IDCAFeeManager.sol';
 import '../utils/Governable.sol';
 
-contract DCAFeeManager is Governable, IDCAFeeManager {
+contract DCAFeeManager is Governable, Multicall, IDCAFeeManager {
+  using SafeERC20 for IERC20;
+
   /// @inheritdoc IDCAFeeManager
   uint16 public constant MAX_TOKEN_TOTAL_SHARE = 10000;
   /// @inheritdoc IDCAFeeManager
   uint32 public constant SWAP_INTERVAL = 1 days;
   /// @inheritdoc IDCAFeeManager
   IDCAHub public immutable hub;
+  /// @inheritdoc IDCAFeeManager
+  IWrappedProtocolToken public immutable wToken;
+  /// @inheritdoc IDCAFeeManager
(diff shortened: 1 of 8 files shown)


contracts/DCAFeeManager/DCAFeeManager.sol
```diff
@@ -12,7 +12,50 @@ contract DCAFeeManager is Governable, IDCAFeeManager {
   /// @inheritdoc IDCAFeeManager
   IDCAHub public immutable hub;
 
-  constructor(IDCAHub _hub, address _governor) Governable(_governor) {
+  TargetTokenShare[] internal _distribution;
+
+  constructor(
+    IDCAHub _hub,
+    TargetTokenShare[] memory _distributionToSet,
```
0xged, line 19: is `calldata` not allowed in constructor?

contracts/DCAFeeManager/DCAFeeManager.sol
```diff
@@ -12,7 +12,50 @@ contract DCAFeeManager is Governable, IDCAFeeManager {
...
+    address _governor
+  ) Governable(_governor) {
     hub = _hub;
+    setTargetTokensDistribution(_distributionToSet);
+  }
+
+  /// @inheritdoc IDCAFeeManager
+  function targetTokensDistribution() external view returns (TargetTokenShare[] memory) {
+    return _distribution;
+  }
```
0xged, line 29: it didnt allow you to make the variable public? 😢

contracts/DCAFeeManager/DCAFeeManager.sol
```diff
@@ -12,7 +12,50 @@ contract DCAFeeManager is Governable, IDCAFeeManager {
...
+      for (uint256 i = _min; i < _newDistribution.length; i++) {
+        _assignedShares += _newDistribution[i].shares;
+        _distribution.push(_newDistribution[i]);
+      }
+    } else if (_currentTargetTokens > _newDistribution.length) {
+      // If have less tokens than before, then remove extra tokens
+      for (uint256 i = _min; i < _currentTargetTokens; i++) {
+        _distribution.pop();
+      }
+    }
```
0xged, line 56: I think you can avoid re-declaring `i` here. If you declare `i` out of the first `for` scope, then at the end of the loop `i = min`, right? and you can avoid the re-declaration.

0xged (approved):
other than the comments, that are nit, this looks good!

contracts/DCAFeeManager/DCAFeeManager.sol
nchamo: No, must be memory. Not sure why

contracts/DCAFeeManager/DCAFeeManager.sol
nchamo: I can make it public, but the getter will be by index, not the entire array. That's why I did it like this

contracts/DCAFeeManager/DCAFeeManager.sol
nchamo: Great catch!

PR #78, opened 2022-05-21
feat: add access management

nchamo:
We are now adding access management. The contract owner/governor will be able to give other users the ability to set the target tokens distribution, and also make withdraws (in future PRs)


0xged (approved):
Looks good. I wonder tho, why not just go with OZ roles? I guess because there might be only one role here?

Other than that, lg2m! Nice work!

nchamo:
> Looks good. I wonder tho, why not just go with OZ roles? I guess because there might be only one role here?
> 
> Other than that, lg2m! Nice work!

Exactly. I thought that OZ roles would be kind of an overkill in this case. A mapping was more than enough, and in my opinion, more easy to use/check

PR #79, opened 2022-05-22
feat: add withdraw protocol token

nchamo:
We are now adding `withdrawProtocolToken`. This will allow the governor or allowed users to withdraw all wToken from the hub & fees, convert it to the protocol token, and then send it to the recipient


test/unit/DCAFeeManager/dca-fee-manager.spec.ts
```diff
@@ -48,6 +61,62 @@ contract('DCAFeeManager', () => {
...
+      given(async () => {
+        await DCAFeeManager.connect(governor).withdrawProtocolToken([], RECIPIENT);
+      });
+      then('no withdraw is executed', () => {
+        expect(DCAHub.withdrawSwappedMany).to.have.not.have.been.called;
+      });
+    });
+    when('all sources have balance', () => {
+      const PLATFORM_BALANCE = utils.parseEther('1');
+      const POSITION_IDS = [1, 2, 3].map(BigNumber.from);
```
0xged, line 91: lol

test/unit/DCAFeeManager/dca-fee-manager.spec.ts
```diff
@@ -212,4 +287,13 @@ contract('DCAFeeManager', () => {
...
   }
+
+  function getPlatformBalance(address: string) {
+    return ethers.provider.getBalance(address);
+  }
+
+  async function setPlatformTokenBalance(recipient: { address: string }, amount: BigNumber) {
+    await ethers.provider.send('hardhat_setBalance', [recipient.address, ethers.utils.hexValue(amount)]);
+    return BigNumber.from(amount);
+  }
```
0xged, line 298: To be completely honest `platformBalance` here is pretty confusing. Since that is what we call the fees on the `Hub`.

0xged (approved):
Ok, not completely in love on how `withdrawProtocolToken` looks tested, but I can not find a better way to just test out the basics. This combined with some e2e might help me feel more comfortable.

test/unit/DCAFeeManager/dca-fee-manager.spec.ts
nchamo: You are right, it should be setProtocolTokenBalance. Platform balance is the fee on the hub. Protocol token is the chain's tokens

For v3 we should rethink names haha

contracts/DCAFeeManager/DCAFeeManager.sol
```diff
@@ -1,33 +1,118 @@
...
   /// @inheritdoc IDCAFeeManager
   IDCAHub public immutable hub;
   /// @inheritdoc IDCAFeeManager
+  IWrappedProtocolToken public immutable wToken;
+  /// @inheritdoc IDCAFeeManager
   mapping(address => bool) public hasAccess;
-
-  TargetTokenShare[] internal _distribution;
+  /// @inheritdoc IDCAFeeManager
+  mapping(bytes32 => uint256) public positions; // key(from, to) => position id
```
0xged, line 24: I don't understand how / why this works without overriding. I'm really confused here. It applies to all stuff on the contract.

contracts/DCAFeeManager/DCAFeeManager.sol
```diff
@@ -1,33 +1,118 @@
...
+      uint256 _amountSpent;
+      for (uint256 j; j < _distribution.length; j++) {
+        uint256 _amountToDeposit = j < _distribution.length - 1
+          ? (_amount.amount * _distribution[j].shares) / MAX_TOKEN_TOTAL_SHARE
+          : _amount.amount - _amountSpent; // If this is the last token, then assign everything that hasn't been spent. We do this to prevent unspent tokens due to rounding errors
+
+        bool _failed = _depositToHub(_amount.token, _distribution[j].token, _amountToDeposit, _amount.amountOfSwaps);
+        if (!_failed) {
+          _amountSpent += _amountToDeposit;
+        }
```
0xged, line 103: Should we revert here? Instead of try/catching. Try/catching and doing this would lead to unintended behavior. Obviously, its not big deal since its our fees... But, why should we keep going if something failed?

contracts/DCAFeeManager/DCAFeeManager.sol
```diff
@@ -39,40 +124,75 @@ contract DCAFeeManager is Governable, IDCAFeeManager {
...
+    for (uint256 i; i < _positionIds.length; i++) {
+      IDCAHubPositionHandler.UserPosition memory _userPosition = hub.userPosition(_positionIds[i]);
+      _balances[i] = PositionBalance({
+        positionId: _positionIds[i],
+        from: _userPosition.from,
+        to: _userPosition.to,
+        swappedBalance: _userPosition.swapped
+      });
     }
+  }
```
0xged, line 156: Why not keep an array of created positions and also return that on `positionBalances` ?

contracts/DCAFeeManager/DCAFeeManager.sol
```diff
@@ -39,40 +124,75 @@ contract DCAFeeManager is Governable, IDCAFeeManager {
...
       }
     }
-
-    if (_assignedShares != MAX_TOKEN_TOTAL_SHARE) revert InvalidAmountOfShares();
-    emit NewDistribution(_newDistribution);
   }
 
   modifier onlyOwnerOrAllowed() {
-    if (!isGovernor(msg.sender) && !hasAccess[msg.sender]) revert CallerMustBeOwnerOrHaveAccess();
+    if (!hasAccess[msg.sender] && !isGovernor(msg.sender)) revert CallerMustBeOwnerOrHaveAccess();
```
0xged, line 195: hmkei 😂

0xged (commented):
The implementation looks good, and I couldn't find breaking points. Although it became almost a monolithic PR (fuck github). 

I'm not entirely sold on this `FeeManager` workflow tho. Please correct me if I'm wrong but currently the workflow would be:

- Withdraw platform balances and fees to `FeeManager` [With off-chain computing]
- Fill positions `from` each of those tokens to another (might be more than one) `to` [With off-chain computing] 
- Withdraw balance from positions (probably on platform token: eth, matic, whatever). [With off-chain computing]

As far as I'm seeing it this creates more overhead than
- Withdraw everyhing from platform to msig, or EOA
- Submit trades for all tokens => platform token

 Even the off-chain workload seems lower. I feel that I'm missing something tho, so please help me to fill in the gaps I might be having.

contracts/DCAFeeManager/DCAFeeManager.sol
nchamo: Done in https://github.com/Mean-Finance/dca-v2-periphery/pull/88

PR #80, opened 2022-05-22
refactor: remove token distribution

nchamo:
We are basically undoing #77. I realized that it would be better if the action to take fees and send fill positions was permissioned, so we can just take the distribution as a parameter, instead of having it stored

My bad 😅


0xged (approved):
At this point in time I've got no idea what `fillPositions` is, but I'm gonna trust yo and see #81 lmao.

PR #82, opened 2022-05-22
feat: add withdraws

nchamo:
We are now adding a few different methods to be able to withdraw funds from the platform balance, the contract itself or positions

### DIFF
--- a/contracts/DCAFeeManager/DCAFeeManager.sol
+++ b/contracts/DCAFeeManager/DCAFeeManager.sol
@@ -1,10 +1,14 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity >=0.8.7 <0.9.0;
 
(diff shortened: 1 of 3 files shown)


contracts/DCAFeeManager/DCAFeeManager.sol
```diff
@@ -122,7 +143,7 @@ contract DCAFeeManager is Governable, IDCAFeeManager {
   }
 
   modifier onlyOwnerOrAllowed() {
-    if (!isGovernor(msg.sender) && !hasAccess[msg.sender]) revert CallerMustBeOwnerOrHaveAccess();
+    if (!hasAccess[msg.sender] && !isGovernor(msg.sender)) revert CallerMustBeOwnerOrHaveAccess();
```
nchamo, line 146: I figured that permissioned actions would be mostly triggered by allowed addresses, and not the governor. So by chaning the order, we can avoid a SLOAD

PR #88, opened 2022-05-25
feat: store position ids

nchamo:
We are now making a small change where we store all created positions in an array. This will allow us to return all position when `availableBalances` is called, and therefore there will be no need for an off-chain way to find the positions.

This should make the entire UX much easier to use.

Also, since we can return everything in `availableBalances`, we removed `positionBalances`


0xged (approved):
Looks good! Loving the changes. Great work 🚀

PR #89, opened 2022-07-05
refactor: split protocol token

nchamo:
Before this change, we had `withdrawProtocolToken`. This function would combine 
1. `withdrawFromPlatformBalance`
2. `withdrawFromPositions` 
3. Unwrap the wToken 
4. Transfer the protocol token (ETH, MATIC, etc) to the recipient.

By using the fee manager, we realized that:
1. Since we have `Multicall`, there is no need to combine 1 and 2 in another function
2. It would be nice to be able to have 4 separetely, specially if someone (👋 ) is stupid enough to send ETH to the contract

So we are now deleting `withdrawProtocolToken` and creating separate functions for (3) and (4)


.github/workflows/gas-report.yml
```diff
@@ -1,49 +0,0 @@
-name: Gas report
```
nchamo, line 1: Deleting gas report, because it's not working

0xged (approved):
Looks great! I'm not sure how will be the process to deprecate the old fee manager and use this one tho. I can help the migration but not sure about all the details. 
P.S: Remember to update deployment script salt 🧂

PR #91, opened 2022-07-23
refactor: take hub address withdraw platform balance

nchamo:
The Fee Manager currently takes a DCAHub address on the constructor. Since it's immutable, it cannot be changed. We could make it mutable, but that would make all actions more expensive

The idea is to add the hub as a parameter to all actions, so that the same Fee Manager can work for multiple hubs at the same time


0xged (approved):
Looks good - I wonder where do you think the positions to get swapped should be, or how should we handle that. I guess its in upcoming PRs tho, so this is a change in a good directon.

PR #92, opened 2022-07-23
refactor: handle allowance reset internally

nchamo:
We are now deleting `resetAllowance` and handling everything internally. It didn't make much sense to have it in another function, since `fillPositions` would revert until `resetAllowance` was called


0xged (approved):
I guess this was a gas optimization back then - feels like a good trade off on our workload tho. 
A little bit more gas consumption, less burden on our side, I'll take it.

PR #93, opened 2022-07-23
refactor: take hub address withdraw from positions

nchamo:
We are now taking the Hub as a parameter when calling `withdrawFromPositions`


0xged (approved):
Send it! 🚀

PR #95, opened 2022-07-23
refactor: take hub address terminate positions

nchamo:
We are now taking the Hub as a parameter when calling `terminatePositions`


0xged (approved):
Looks good to go!

PR #97, opened 2022-07-23
refactor: stop taking hub as contract parameter

nchamo:
After all the recent PRs, we can now stop taking the hub as a contract parameter

### DIFF
--- a/deploy/005_fee_manager.ts
+++ b/deploy/005_fee_manager.ts
@@ -34,8 +34,6 @@ const deployFunction: DeployFunction = async function (hre: HardhatRuntimeEnviro
       throw new Error(`Unsupported chain '${hre.network.name}`);
   }
 
-  const hub = await hre.deployments.get('DCAHub');
-
   await deployThroughDeterministicFactory({
     deployer,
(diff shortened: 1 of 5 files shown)


deploy/005_fee_manager.ts
```diff
@@ -34,17 +34,15 @@ const deployFunction: DeployFunction = async function (hre: HardhatRuntimeEnviro
       throw new Error(`Unsupported chain '${hre.network.name}`);
   }
 
-  const hub = await hre.deployments.get('DCAHub');
-
   await deployThroughDeterministicFactory({
     deployer,
     name: 'DCAFeeManager',
     salt: 'MF-DCAV2-DCAFeeManager-V2',
```
0xged, line 40: We had already updated this, right?

0xged (approved):
Noice! Multiple hub fee manager! 🚀

deploy/005_fee_manager.ts
nchamo: Maybe? Haha. We will know for sure when we try to deploy

PR #98, opened 2022-07-23
refactor: change governable for access control

nchamo:
We are now simplifying the Fee Manager. We had a governor + allowed. This was basically AccessControl with two different roles. For consistency, we are moving towards AccessControl + Super Admin + Admins, like in our other repos

Look at all that deleted code
image


0xged (approved):
Hahaha! Love it. I'm pretty sure I did say "hey, why not access control?" - But I'm glad we are finally moving into this. A bit more costly in general but I'd rather have this as a standard. 
Great work!

PR #99, opened 2022-07-26
refactor: remove collectable dust and governor

0xsambugs:
We are now removing `CollectableDust` and `Governable` for the Companion


0xged (commented):
I'm guessing the idea here will be to inherit the `swappers/extensions` - So I'm going to go ahead and approve this PR. 
But if we are not gonna inherit that stuff from extensions: Why? Seems wise to have collectable dust.

0xged (approved):

0xsambugs:
> I'm guessing the idea here will be to inherit the `swappers/extensions` - So I'm going to go ahead and approve this PR. But if we are not gonna inherit that stuff from extensions: Why? Seems wise to have collectable dust.

Yes, we are going to inherit everything from the `swappers/extensions`, that's why are removing the code here

nchamo (approved):
Looks good, but next time please clarify that these functionalities will be re-added later 😅

PR #100, opened 2022-07-26
refactor: avoid taking tokens from caller

0xsambugs:
We continue with small changes. Right now, we are no longer taking tokens from the caller when executing `depositProxy` or `increasePositionProxy`


0xged (approved):
I'm guessing this is part of a general refactor, since taking transfer functionality seems a bit off. 
But, in terms of the change, and the tests - they look alright!

0xsambugs:
> I'm guessing this is part of a general refactor, since taking transfer functionality seems a bit off. But, in terms of the change, and the tests - they look alright!

Yes, I did a very shitty job explaining this haha. The idea is to make each function as small and independent as possible. I will add more functions that will take tokens from the caller, and the idea is that clients can combine them using multicall

nchamo (approved):

PR #101, opened 2022-07-27
refactor: add hub as parameter

0xsambugs:
We are now taking the `DCAHub` and the `DCAPermissionManager` as parameters on the Companion. Also, we changed the docs so that they were consistent with are new repositories

### DIFF
--- a/test/unit/DCAHubCompanion/dca-hub-companion-multi-call-handler.spec.ts
+++ b/test/unit/DCAHubCompanion/dca-hub-companion-multi-call-handler.spec.ts
@@ -32,6 +32,7 @@ contract('DCAHubCompanionMulticallHandler', () => {
     );
     DCAPermissionManager = await smock.fake('IDCAPermissionManager');
     DCAHub = await smock.fake('IDCAHub');
+    DCAHub.permissionManager.returns(DCAPermissionManager.address);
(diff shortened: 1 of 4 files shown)


test/unit/DCAHubCompanion/dca-hub-companion-multi-call-handler.spec.ts
```diff
@@ -247,6 +278,7 @@ contract('DCAHubCompanionMulticallHandler', () => {
     params: [10, constants.NOT_ZERO_ADDRESS, constants.NOT_ZERO_ADDRESS],
   });
 
+  type DropFirstParam<T extends unknown[]> = T extends [any, ...infer U] ? U : never;
```
0xged, line 281: Some voodoo sheit here.

0xged (approved):
Looks good! This is becoming interesting.

test/unit/DCAHubCompanion/dca-hub-companion-multi-call-handler.spec.ts
0xsambugs: I googled this haha

nchamo (approved):
Thanks for adding consistency to the docs 😊

PR #104, opened 2022-07-27
feat: deposit with all balance

0xsambugs:
One of the new use cases that we want to enable is:
1. Take token from user
2. Convert token to another token (maybe the yield-bearing version of the original token)
3. Deposit the converted token to the Hub

Since slippage can happen in the conversion, the idea is to have a function that takes the Companion's entire balance

### DIFF
--- a/contracts/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
+++ b/contracts/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
@@ -32,11 +32,26 @@ abstract contract DCAHubCompanionHubProxyHandler is IDCAHubCompanionHubProxyHand
     address _owner,
     IDCAPermissionManager.PermissionSet[] calldata _permissions,
     bytes calldata _miscellaneous
-  ) external payable returns (uint256 _positionId) {
+  ) public payable virtual returns (uint256 _positionId) {
     _approveHub(address(_from), _hub, _amount);
     _positionId = _miscellaneous.length > 0
       ? _hub.deposit(_from, _to, _amount, _amountOfSwaps, _swapInterval, _owner, _permissions, _miscellaneous)
       : _hub.deposit(_from, _to, _amount, _amountOfSwaps, _swapInterval, _owner, _permissions);
+  }
+
+  /// @inheritdoc IDCAHubCompanionHubProxyHandler
+  function depositWithAllBalanceProxy(
+    IDCAHub _hub,
+    address _from,
+    address _to,
+    uint32 _amountOfSwaps,
+    uint32 _swapInterval,
+    address _owner,
+    IDCAPermissionManager.PermissionSet[] calldata _permissions,
+    bytes calldata _miscellaneous
+  ) external payable returns (uint256 _positionId) {
+    uint256 _amount = IERC20(_from).balanceOf(address(this));
+    return depositProxy(_hub, _from, _to, _amount, _amountOfSwaps, _swapInterval, _owner, _permissions, _miscellaneous);
   }
 
   /// @inheritdoc IDCAHubCompanionHubProxyHandler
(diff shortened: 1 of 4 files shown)


contracts/mocks/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
```diff
@@ -3,4 +3,49 @@ pragma solidity >=0.8.7 <0.9.0;
 
 import '../../DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol';
 
-contract DCAHubCompanionHubProxyHandlerMock is DCAHubCompanionHubProxyHandler {}
+contract DCAHubCompanionHubProxyHandlerMock is DCAHubCompanionHubProxyHandler {
```
0xsambugs, line 6: Tried it with smock, but it didn't detect the internal call (even though the function is public) 😢

contracts/mocks/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
```diff
@@ -3,4 +3,49 @@ pragma solidity >=0.8.7 <0.9.0;
...
+    address _to,
+    uint256 _amount,
+    uint32 _amountOfSwaps,
+    uint32 _swapInterval,
+    address _owner,
+    IDCAPermissionManager.PermissionSet[] calldata _permissions,
+    bytes calldata _miscellaneous
+  ) public payable override returns (uint256 _positionId) {
+    _depositProxyCalls.push();
+    DepositProxyCall storage _ref = _depositProxyCalls[_depositProxyCalls.length - 1];
```
nchamo, line 37: Is there a reason why you split it like this?

nchamo (approved):
Looks good, just have one question

contracts/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
```diff
@@ -32,13 +32,28 @@ abstract contract DCAHubCompanionHubProxyHandler is IDCAHubCompanionHubProxyHand
...
-  ) external payable returns (uint256 _positionId) {
+  ) public payable virtual returns (uint256 _positionId) {
     _approveHub(address(_from), _hub, _amount);
     _positionId = _miscellaneous.length > 0
       ? _hub.deposit(_from, _to, _amount, _amountOfSwaps, _swapInterval, _owner, _permissions, _miscellaneous)
       : _hub.deposit(_from, _to, _amount, _amountOfSwaps, _swapInterval, _owner, _permissions);
   }
 
+  /// @inheritdoc IDCAHubCompanionHubProxyHandler
+  function depositWithAllBalanceProxy(
```
0xged, line 43: Do we need to keep adding `Proxy` on the functions name? I think we can take 'em out, right?

contracts/mocks/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
0xged: +++ I wanna know! 😂

0xged (approved):
Looks good, can't wait to get this audited 🤓 xd

contracts/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
0xsambugs: All of them, or just the new ones?

contracts/mocks/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
0xsambugs: Hahah. I tried to push it directly, but Solidity doesn't support array conversions from calldata to storage yet, so I had to do it manually

contracts/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
0xged: I'd take the `Proxy` suffix out of all the functions.

0xsambugs: Will do it in #105

PR #105, opened 2022-07-27
feat: increase with all balance

0xsambugs:
Some as #104, but with `increasePosition` instead of `deposit`

### DIFF
--- a/contracts/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
+++ b/contracts/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
@@ -9,7 +9,7 @@ abstract contract DCAHubCompanionHubProxyHandler is IDCAHubCompanionHubProxyHand
   using SafeERC20 for IERC20Metadata;
 
   /// @inheritdoc IDCAHubCompanionHubProxyHandler
-  function permissionPermitProxy(
+  function permissionPermit(
     IDCAPermissionManager _permissionManager,
     IDCAPermissionManager.PermissionSet[] calldata _permissions,
     uint256 _tokenId,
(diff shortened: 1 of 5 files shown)


nchamo (approved):
Looks good!

contracts/DCAHubCompanion/DCAHubCompanionHubProxyHandler.sol
```diff
@@ -89,6 +89,18 @@ abstract contract DCAHubCompanionHubProxyHandler is IDCAHubCompanionHubProxyHand
     _hub.increasePosition(_positionId, _amount, _newSwaps);
   }
 
+  /// @inheritdoc IDCAHubCompanionHubProxyHandler
+  function increasePositionWithAllBalanceProxy(
```
0xged, line 93: I'm thinking name should probably be `increasePOsitionWithAlllCompanionsBalance`? Or something along those lines? Because with `AllBalance` it seems that we are talking about the user's balance.

0xged: Also: why not re-use the `increasePositionProxy` code?

0xsambugs: Just to optimize gas. I would need to call `_hub.userPosition(_positionId).from;` twice (once here and once inside `increasePositionProxy`). I though about creating an internal function that did approve + increase, but it really didn't make much sense to refactor everything for two lines of code

0xged: No, for sure. Refactoring everything doesn't make sense. Just that it adds overhead on testing (since you can't just test the "internal" call. But its ok then.

0xged (approved):

PR #106, opened 2022-07-27
feat: take withdraw and swap handler

0xsambugs:
We are now adding a new handler called `DCAHubCompanionTakeWithdrawAndSwapHandler`.

This handler will to the Companion the ability to:
- Take tokens from the `msg.sender`
- Withdraw tokens form the contract and send them to a given recipient
- Execute swaps

## Why are we adding this?
This handler, together will multicall, will enable all kinds of interesting behaviours, such as:
- Take 1 ETH from the user, swap it to WETH, and create two different positions, each with 0.5 WETH
- Withdraw all swapped tokens from a position, swap it for another, and transfer them to a given address
- Take funds from one DCAHub, swap the funds for another token, and deposit it into another DCAHub

This type of flexibility will enable use cases that we can't even think of right now. It will also allow us to migrate positions from one DCAHub to the other, so we can remove the specific contract that we've built for it

### DIFF
--- a/contracts/DCAHubCompanion/DCAHubCompanionTakeWithdrawAndSwapHandler.sol
+++ b/contracts/DCAHubCompanion/DCAHubCompanionTakeWithdrawAndSwapHandler.sol
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.7 <0.9.0;
+
+import '@mean-finance/swappers/solidity/contracts/SwapAdapter.sol';
+import '../interfaces/IDCAHubCompanion.sol';
+
+/// @dev All public functions are payable, so that they can be multicalled together with other payable functions when msg.value > 0
+abstract contract DCAHubCompanionTakeWithdrawAndSwapHandler is SwapAdapter, IDCAHubCompanionTakeWithdrawAndSwapHandler {
+  using SafeERC20 for IERC20;
+  using Address for address payable;
+
+  /// @inheritdoc IDCAHubCompanionTakeWithdrawAndSwapHandler
+  function sendToRecipient(
+    address _token,
+    uint256 _amount,
+    address _recipient
+  ) external payable {
+    if (_token == PROTOCOL_TOKEN) {
+      payable(_recipient).sendValue(_amount);
+    } else {
+      IERC20(_token).safeTransfer(_recipient, _amount);
+    }
+  }
+
+  /// @inheritdoc IDCAHubCompanionTakeWithdrawAndSwapHandler
+  function takeFromCaller(IERC20 _token, uint256 _amount) external payable {
+    _takeFromMsgSender(_token, _amount);
+  }
+
+  /// @inheritdoc IDCAHubCompanionTakeWithdrawAndSwapHandler
+  function sendAllBalanceToRecipient(address _token, address _recipient) external payable {
+    _sendBalanceToRecipient(_token, _recipient);
+  }
+}
--- a/contracts/interfaces/IDCAHubCompanion.sol
+++ b/contracts/interfaces/IDCAHubCompanion.sol
@@ -59,6 +59,9 @@ interface IDCAHubCompanionTakeSendAndSwapHandler {
 }
 
 interface IDCAHubCompanionHubProxyHandler {
+  /// @notice Thrown when a user tries operate on a position that they don't have access to
+  error UnauthorizedCaller();
+
   /**
    * @notice Creates a new position
    * @dev Meant to be used as part of a multicall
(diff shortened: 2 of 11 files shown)


contracts/DCAHubCompanion/DCAHubCompanionTakeWithdrawAndSwapHandler.sol
```diff
@@ -0,0 +1,34 @@
...
+  ) external payable {
+    if (_token == PROTOCOL_TOKEN) {
+      payable(_recipient).sendValue(_amount);
+    } else {
+      IERC20(_token).safeTransfer(_recipient, _amount);
+    }
+  }
+
+  /// @inheritdoc IDCAHubCompanionTakeWithdrawAndSwapHandler
+  function takeFromCaller(IERC20 _token, uint256 _amount) external payable {
```
0xsambugs, line 26: I didn't add tests for `takeFromCaller` and `sendAllBalanceToRecipient`, since they are only exposing already tested functions. Is that ok for you?

contracts/DCAHubCompanion/DCAHubCompanionTakeWithdrawAndSwapHandler.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.7 <0.9.0;
+
+import '@mean-finance/swappers/solidity/contracts/SwapAdapter.sol';
+import '../interfaces/IDCAHubCompanion.sol';
+
+/// @dev All public functions are payable, so that they can be multicalled together with other payable functions when msg.value > 0
+abstract contract DCAHubCompanionTakeWithdrawAndSwapHandler is SwapAdapter, IDCAHubCompanionTakeWithdrawAndSwapHandler {
```
0xsambugs, line 8: In a following PR, I'll Swap Extensions to the contract

nchamo: I'm assuming you will add the extensions? Hehe

contracts/interfaces/IDCAHubCompanion.sol
```diff
@@ -192,10 +221,7 @@ interface IDCAHubCompanionHubProxyHandler {
   ) external payable;
 }
 
-interface IDCAHubCompanion is IDCAHubCompanionLibrariesHandler, IDCAHubCompanionHubProxyHandler {
-  /// @notice Thrown when one of the parameters is a zero address
-  error ZeroAddress();
-
+interface IDCAHubCompanion is IDCAHubCompanionLibrariesHandler, IDCAHubCompanionHubProxyHandler, IDCAHubCompanionTakeWithdrawAndSwapHandler {
   /// @notice Thrown when a user tries operate on a position that they don't have access to
   error UnauthorizedCaller();
```
nchamo, line 226: Does this need to be here? Or could we move this to `IDCAHubCompanionHubProxyHandler`?

nchamo (approved):
Looks good!

contracts/interfaces/IDCAHubCompanion.sol
```diff
@@ -29,6 +29,35 @@ interface IDCAHubCompanionLibrariesHandler {
   function secondsUntilNextSwap(IDCAHub hub, Pair[] calldata pairs) external view returns (uint256[] memory);
 }
 
+interface IDCAHubCompanionTakeWithdrawAndSwapHandler {
```
0xged, line 32: Why the name `TakeWithdrawAndSwap`? What is it withdrawing? What is it swapping? There is something missing, or this is not the right naming.

contracts/DCAHubCompanion/DCAHubCompanionTakeWithdrawAndSwapHandler.sol
0xged: I mean, you could test the "internal call". It seems kind of "too much" but you know how much I like reducing errors scopes and increased testing lol. Up to you tb completely honest.

contracts/interfaces/IDCAHubCompanion.sol
0xsambugs: Initially, `sendToRecipient` was called `Withdraw`, that's why. Also, I'll be adding Swap extensions to the contract, that the "Swap" part

Would you prefer I call it `TakeSendAndSwapHandler`?

0xged: Oh. Ok. Makes sense now haha. Sure. I'd take your `TakeSendAndSwapHandler`

0xged (approved):

contracts/DCAHubCompanion/DCAHubCompanionTakeWithdrawAndSwapHandler.sol
0xsambugs: Ok, will add the tests

PR #108, opened 2022-07-28
refactor: dca swapper take hub as parameter

0xsambugs:
We are now making two big changes to the `swapForCaller` function

1. We are now allowing ETH to be used instead of WETH any more
Before this change, the caller could send ETH so that it would be used when the hub needed WETH. We are now simplifying the swap by removing this. If the user would want to do this, then they can use the SwapProxy and chain the ETH => WETH swap, together with swapForCaller
2. We are now taking the DCAHub as a parameter. This implies that we won't know the address of the DCAHub, and we will have to assume that someone can specify whatever address they want. Therefore, the one who calls DCAHubSwapCall could be anyone. That's why we are now storing the address who initiated the swap internally


0xged (approved):
This is a great initiative. Although its kinda scary, as discussed offline I think we should be safe. 
Great work Sam!

nchamo (approved):
Love simplifying PRs <3

PR #111, opened 2022-07-29
chore: delete keep3r

0xsambugs:
We are now deleting everything related to keep3r. We don't have any plans to use it anytime soon, and I was tired of having to keep editing its tests so that it kept working

We can re-add it when we build the integration for it


0xged (approved):
😢  but yeah - makes sense.

nchamo (approved):
Look at all that deleted code!

PR #113, opened 2022-07-30
feat: add swap extensions to companion

0xsambugs:
We doing a few things in this PR:
1. We are using the internal function for `sendToRecipient`
2. We are adding the following extensions to the companion:
    - `RunSwap`: so we can execute swaps
    - `RevokableWithGovernor`: to revoke potential access
    - `GetBalances `: to be able to easily read the contract's balance
3. Also, we fixed the Companion's deployment that was working on the tests for some reason 😆

### DIFF
--- a/contracts/DCAHubCompanion/DCAHubCompanionTakeSendAndSwapHandler.sol
+++ b/contracts/DCAHubCompanion/DCAHubCompanionTakeSendAndSwapHandler.sol
@@ -1,11 +1,11 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity >=0.8.7 <0.9.0;
 
-import '@mean-finance/swappers/solidity/contracts/SwapAdapter.sol';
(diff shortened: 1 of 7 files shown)


contracts/DCAHubCompanion/DCAHubCompanionTakeSendAndSwapHandler.sol
```diff
@@ -29,6 +22,6 @@ abstract contract DCAHubCompanionTakeSendAndSwapHandler is SwapAdapter, IDCAHubC
 
   /// @inheritdoc IDCAHubCompanionTakeSendAndSwapHandler
   function sendBalanceOnContractToRecipient(address _token, address _recipient) external payable {
-    _sendBalanceToRecipient(_token, _recipient);
+    _sendBalanceOnContractToRecipient(_token, _recipient);
```
0xsambugs, line 25: We renamed it in the last version

0xged (approved):

nchamo (approved):

PR #116, opened 2022-08-02
feat: make dca swapper permissioned

0xsambugs:
We are now making it so that swaps in the DCASwapper can only be executed by addressed with the `SWAP_EXECUTION_ROLE` role

### DIFF
--- a/contracts/DCAHubSwapper/DCAHubSwapperSwapHandler.sol
+++ b/contracts/DCAHubSwapper/DCAHubSwapperSwapHandler.sol
@@ -66,7 +66,7 @@ abstract contract DCAHubSwapperSwapHandler is DeadlineValidation, AccessControl,
     uint256[] calldata _maximumInput,
     address _recipient,
     uint256 _deadline
-  ) external payable checkDeadline(_deadline) returns (IDCAHub.SwapInfo memory _swapInfo) {
+  ) external payable checkDeadline(_deadline) onlyRole(SWAP_EXECUTION_ROLE) returns (IDCAHub.SwapInfo memory _swapInfo) {
     // Set the swap's executor
     _swapExecutor = msg.sender;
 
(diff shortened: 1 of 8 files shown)


contracts/DCAHubSwapper/DCAHubSwapperSwapHandler.sol
```diff
@@ -96,12 +96,22 @@ abstract contract DCAHubSwapperSwapHandler is DeadlineValidation, AccessControl,
...
   }
 
   /// @inheritdoc IDCAHubSwapperSwapHandler
-  function swapWithDexesForMean(SwapWithDexesParams calldata _parameters) external payable returns (IDCAHub.SwapInfo memory) {
+  function swapWithDexesForMean(SwapWithDexesParams calldata _parameters)
+    external
+    payable
+    onlyRole(SWAP_EXECUTION_ROLE)
+    returns (IDCAHub.SwapInfo memory)
+  {
```
0xged, line 114: If we are doing this permissioned, is it worth having the "non-for-mean"? 👀

0xged (approved):

contracts/DCAHubSwapper/DCAHubSwapperSwapHandler.sol
0xsambugs: I think so. We might allow non-mean users to use this in the future, and it's already coded haha

0xged: Oh. You mean giving auth to someone that is not mean? Like an external MM for example?

nchamo: If that's what you mean Sam, then I guess it makes sense

nchamo (approved):

PR #117, opened 2022-08-02
refactor: remove handler

0xsambugs:
The `DCAHubSwapper` was only composed of the `DCAHubSwapperSwapHandler`, so we removed the `DCAHubSwapperSwapHandler` and moved all the code directly to the `DCAHubSwapper`


0xged (approved):
Awesome :) - simplified.

nchamo (approved):
Makes sense 👍

PR #118, opened 2022-08-02
refactor: change swappers permissions

0xsambugs:
We are now making a small change. We are bringing back the "super admin role" so now everything is consistent with the rest of the permissioned contracts, and we are setting it as its own admin

### DIFF
--- a/contracts/DCAHubSwapper/DCAHubSwapper.sol
+++ b/contracts/DCAHubSwapper/DCAHubSwapper.sol
@@ -36,6 +36,7 @@ contract DCAHubSwapper is DeadlineValidation, AccessControl, SwapAdapter, IDCAHu
   using SafeERC20 for IERC20;
   using Address for address;
 
+  bytes32 public constant SUPER_ADMIN_ROLE = keccak256('SUPER_ADMIN_ROLE');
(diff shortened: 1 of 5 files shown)


0xged (commented):
I do understand consistency... But... `ADMIN_ROLE` here becomes completely pointless? We are having a `Governor` to appoint `Semi-governors` which do nothing. And a `Governor` that appoints `executors`. Not sure about this tbh. If you've got strong feelings about it, lets go, but it seems kinda circular.

0xsambugs:
> I do understand consistency... But... `ADMIN_ROLE` here becomes completely pointless? We are having a `Governor` to appoint `Semi-governors` which do nothing. And a `Governor` that appoints `executors`. Not sure about this tbh. If you've got strong feelings about it, lets go, but it seems kinda circular.

Damn, you are reviewing things too fast. The next PR will add a permissioned `sendDust` and `revokeAllowances`. The idea is that admins will have permissions over those

By doing so, we can have the msig on the super admin role, bots on the admin role, and swap executioners on their own role

0xged (approved):
Woops! Sorry hehe. Then this is goo!

nchamo (approved):
I had the same question as @0xged , but I saw your response 👍

PR #121, opened 2022-08-02
refactor: create base companion

0xsambugs:
We are now creating what we call a `BaseCompanion`. Inheriting from this contract will extend the capabilities of all our companion contracts so that they can execute multicalls, swaps, revokes and more. Take, swap, send, everything will be done by this `BaseCompanion`. 

Our current DCAHubCompanion will then have the proxy calls to the DCAHub, and then this added layer of capabilities. In the future, if we want to build similar companions that interact with other contracts (that are not the DCAHub), we will already have this base contract in place


nchamo (approved):
Wonder if someone is going to find this and get the alpha 👀

0xged (commented):
> In the future, if we want to build similar companions that interact with other contracts...

🕵️ α 

It does make the Companion look better also, so great work!

0xged (approved):

PR #122, opened 2022-08-02
refactor: delete migrator

0xsambugs:
One of the coolest things about everything we've done so far, is that now the DCAHubCompanion is truly flexible. We can run migrations (form one hub to the other) directly on it, so we don't need a Migrator contract any more


nchamo (approved):
On one hand, executing migrations is a little more complex now. On the other, I love that we can use the Companion for all these new use-cases

test/integration/DCAHubCompanion/position-migrator.spec.ts
```diff
@@ -110,11 +114,36 @@ contract('PositionMigrator', () => {
...
+          targetHub().address,
+          position.from,
+          position.to,
+          position.swapsLeft,
+          position.swapInterval,
+          positionOwner.address,
+          [],
+          []
+        );
+        const tx = await DCAHubCompanion.multicall([permissionData!, terminateData!, depositData!]);
```
0xged, line 146: That moment when you only need to change this because you've done an amazing test.

0xged (approved):
Nice! This is a good side-effect all of your work Sam! Major props!

PR #125, opened 2022-08-18
feat: add legacy swap for caller

0xsambugs:
We are now adding a way to execute `swapForCaller` for legacy hubs

### DIFF
--- a/contracts/DCAHubSwapper/DCAHubSwapper.sol
+++ b/contracts/DCAHubSwapper/DCAHubSwapper.sol
@@ -102,6 +102,41 @@ contract DCAHubSwapper is DeadlineValidation, AccessControl, GetBalances, IDCAHu
   }
 
   /// @inheritdoc IDCAHubSwapper
+  function legacySwapForCaller(LegacySwapForCallerParams calldata _parameters)
+    external
+    payable
+    checkDeadline(_parameters.deadline)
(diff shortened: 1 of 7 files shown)


contracts/DCAHubSwapper/DCAHubSwapper.sol
```diff
@@ -101,6 +101,41 @@ contract DCAHubSwapper is DeadlineValidation, AccessControl, GetBalances, IDCAHu
...
+
+    // Check that limits were met
+    for (uint256 i; i < _swapInfo.tokens.length; i++) {
+      ILegacyDCAHub.TokenInSwap memory _tokenInSwap = _swapInfo.tokens[i];
+      if (_tokenInSwap.reward < _parameters.minimumOutput[i]) {
+        revert RewardNotEnough();
+      } else if (_tokenInSwap.toProvide > _parameters.maximumInput[i]) {
+        revert ToProvideIsTooMuch();
+      }
+    }
```
0xged, line 133: Might be worth doing the `unchecked i++`? Since it's a cost-sensible function (same for other loops in the contract)

0xged (approved):
🫡 great work king.

contracts/DCAHubSwapper/DCAHubSwapper.sol
0xsambugs: Will do it in #127

PR #128, opened 2022-08-18
refactor: consider privilege on seconds until next swap

0xsambugs:
We are now taking into consideration the new privileged role when calculating `secondsUntilNextSwap`

### DIFF
--- a/test/unit/libraries/seconds-until-next-swap.spec.ts
+++ b/test/unit/libraries/seconds-until-next-swap.spec.ts
@@ -33,98 +33,204 @@ contract('SecondsUntilNextSwap', () => {
 
     when('seconds are calculated', () => {
       given(async () => {
-        await secondsUntilNextSwap['secondsUntilNextSwap(address,address,address)'](DCAHub.address, TOKEN_B, TOKEN_A);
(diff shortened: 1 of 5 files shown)


test/unit/libraries/seconds-until-next-swap.spec.ts
```diff
@@ -33,98 +33,204 @@ contract('SecondsUntilNextSwap', () => {
...
     });
 
     function nextAvailableAt(relativeLastSwappedAt: number, swapInterval: SwapInterval) {
       return BigNumber.from(CURRENT_TIMESTAMP).add(relativeLastSwappedAt).div(swapInterval.seconds).add(1).mul(swapInterval.seconds);
     }
 
+    function unprivilegedNextAvailableAt(relativeLastSwappedAt: number, swapInterval: SwapInterval) {
+      const extra = BigNumber.from(swapInterval.seconds).div(3);
+      return nextAvailableAt(relativeLastSwappedAt, swapInterval).add(extra);
+    }
```
0xged, line 224: Nice!

0xged (approved):
Nice catch. For some reason I thought this leveraged dca-v2-core, completely missed it. Seems like a pretty straight forward testing, and change in solidity is smol and looks good.

PR #129, opened 2022-08-18
refactor: support new params next swap info

0xsambugs:
We are now adding support for the new params in getNextSwapInfo

### DIFF
--- a/contracts/interfaces/ILegacyDCAHub.sol
+++ b/contracts/interfaces/ILegacyDCAHub.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity >=0.8.7 <0.9.0;
 
+import '@mean-finance/dca-v2-core/contracts/interfaces/IDCAHub.sol';
+
 interface ILegacyDCAHub {
   /// @notice Information about a swap
   struct SwapInfo {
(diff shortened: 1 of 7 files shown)


contracts/interfaces/ILegacyDCAHub.sol
```diff
@@ -36,14 +38,6 @@ interface ILegacyDCAHub {
     bytes1 intervalsInSwap;
   }
 
-  /// @notice A pair of tokens, represented by their indexes in an array
-  struct PairIndexes {
```
0xsambugs, line 40: I removed this struct here and I'm using the one from the new DCAHub because Solidity can't tell that they are actually the same. So in order to be able to create a pair that can be used for both the legacy and new hubs, they need to be the same

test/integration/DCAHubSwapper/multi-pair-swap-with-dex.spec.ts
```diff
@@ -58,6 +58,7 @@ contract('Multi pair swap with DEX', () => {
     //We are setting a very high fee, so that there is a surplus in both reward and toProvide tokens
     await DCAHub.connect(timelock).setSwapFee(50000); // 5%
     // Allow swapper
+    await DCAHub.connect(governor).grantRole(await DCAHub.PRIVILEGED_SWAPPER_ROLE(), DCAHubSwapper.address);
```
0xsambugs, line 61: If I didn't add this, the test would be flaky. Depending on how fast the test was executed, it would fail 😅

contracts/interfaces/ILegacyDCAHub.sol
0xged: I was wondering why that change. Good to know.

0xged (approved):
gw!

PR #130, opened 2022-08-19
feat: basic structure

TheAustrian1998:
Create Strategies basic structure:

- create interfaces
- create abstract that inherits from those interfaces
- create main contract that inherits those abstracts
- add nft-descriptor package

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -3,4 +3,24 @@ pragma solidity >=0.8.7 <0.9.0;
 
 import '../../interfaces/IDCAStrategies.sol';
 
-abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {}
+abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
+  function getStrategy(uint80 _strategyId) external view override returns (Strategy memory) {}
+
+  function getStrategyIdByName(string memory _strategyName) external view override returns (uint80 _strategyId) {}
+
+  function createStrategy(
+    string memory _strategyName,
+    ShareOfToken[] memory _tokens,
+    address _owner
+  ) external override returns (uint80 _strategyId) {}
+
+  function updateStrategyTokens(uint80 _strategyId, ShareOfToken[] memory _tokens) external override {}
+
+  function updateStrategyName(uint80 _strategyId, string memory _newStrategyName) external override {}
+
+  function transferStrategyOwnership(uint80 _strategyId, address _newOwner) external override {}
+
+  function acceptStrategyOwnership(uint80 _strategyId) external override {}
+
+  function cancelStrategyOwnershipTransfer(uint80 _strategyId, address _newOwner) external override {}
+}
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -4,23 +4,31 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
+  /// @inheritdoc IDCAStrategiesManagementHandler
   function getStrategy(uint80 _strategyId) external view override returns (Strategy memory) {}
 
+  /// @inheritdoc IDCAStrategiesManagementHandler
   function getStrategyIdByName(string memory _strategyName) external view override returns (uint80 _strategyId) {}
 
+  /// @inheritdoc IDCAStrategiesManagementHandler
   function createStrategy(
     string memory _strategyName,
     ShareOfToken[] memory _tokens,
     address _owner
   ) external override returns (uint80 _strategyId) {}
 
+  /// @inheritdoc IDCAStrategiesManagementHandler
   function updateStrategyTokens(uint80 _strategyId, ShareOfToken[] memory _tokens) external override {}
 
+  /// @inheritdoc IDCAStrategiesManagementHandler
   function updateStrategyName(uint80 _strategyId, string memory _newStrategyName) external override {}
 
+  /// @inheritdoc IDCAStrategiesManagementHandler
   function transferStrategyOwnership(uint80 _strategyId, address _newOwner) external override {}
 
+  /// @inheritdoc IDCAStrategiesManagementHandler
   function acceptStrategyOwnership(uint80 _strategyId) external override {}
 
+  /// @inheritdoc IDCAStrategiesManagementHandler
   function cancelStrategyOwnershipTransfer(uint80 _strategyId) external override {}
 }
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -1,0 +1,6 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.7 <0.9.0;
+
+import '../../interfaces/IDCAStrategies.sol';
+
+abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {}
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -5,7 +5,7 @@ import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
   /// @inheritdoc IDCAStrategiesManagementHandler
-  function getStrategy(uint80 _strategyId) external view override returns (Strategy memory) {}
+  function getStrategy(uint80 _strategyId) external view override returns (IDCAStrategies.Strategy memory) {}
 
   /// @inheritdoc IDCAStrategiesManagementHandler
   function getStrategyIdByName(string memory _strategyName) external view override returns (uint80 _strategyId) {}
@@ -13,12 +13,12 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
   /// @inheritdoc IDCAStrategiesManagementHandler
   function createStrategy(
     string memory _strategyName,
-    ShareOfToken[] memory _tokens,
+    IDCAStrategies.ShareOfToken[] memory _tokens,
     address _owner
   ) external override returns (uint80 _strategyId) {}
 
   /// @inheritdoc IDCAStrategiesManagementHandler
-  function updateStrategyTokens(uint80 _strategyId, ShareOfToken[] memory _tokens) external override {}
+  function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external override {}
 
   /// @inheritdoc IDCAStrategiesManagementHandler
   function updateStrategyName(uint80 _strategyId, string memory _newStrategyName) external override {}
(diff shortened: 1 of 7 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -0,0 +1,26 @@
...
+
+  function updateStrategyTokens(uint80 _strategyId, ShareOfToken[] memory _tokens) external override {}
+
+  function updateStrategyName(uint80 _strategyId, string memory _newStrategyName) external override {}
+
+  function transferStrategyOwnership(uint80 _strategyId, address _newOwner) external override {}
+
+  function acceptStrategyOwnership(uint80 _strategyId) external override {}
+
+  function cancelStrategyOwnershipTransfer(uint80 _strategyId, address _newOwner) external override {}
```
nchamo, line 25: Do we need ` address _newOwner` here? There should only be one pending transfer, so I don't think we need the new owner

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
```diff
@@ -0,0 +1,61 @@
...
+  // solhint-disable-next-line func-name-mixedcase
+  function PERMISSION_PERMIT_TYPEHASH() external pure override returns (bytes32) {}
+
+  // solhint-disable-next-line func-name-mixedcase
+  function PERMISSION_SET_TYPEHASH() external pure override returns (bytes32) {}
+
+  // solhint-disable-next-line func-name-mixedcase
+  function DOMAIN_SEPARATOR() external view override returns (bytes32) {}
+
+  function nftDescriptor() external override returns (IDCAHubPositionDescriptor) {}
```
nchamo, line 20: We will use a different descriptor here, so could you add a `TODO` comment that we need to update this when we build said descriptor?

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
```diff
@@ -0,0 +1,61 @@
...
+
+  function hub() external override returns (address) {}
```
nchamo, line 22: Luckily, we don't this here. Each position will have its own hub address

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
```diff
@@ -0,0 +1,61 @@
...
+    Permission _permission
+  ) external view override returns (bool) {}
+
+  function hasPermissions(
+    uint256 _id,
+    address _account,
+    Permission[] calldata _permissions
+  ) external view override returns (bool[] memory _hasPermissions) {}
+
+  function setHub(address _hub) external override {}
```
nchamo, line 38: Same here, we don't need this

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
```diff
@@ -0,0 +1,61 @@
...
+  function permissionPermit(
+    PermissionSet[] calldata _permissions,
+    uint256 _tokenId,
+    uint256 _deadline,
+    uint8 _v,
+    bytes32 _r,
+    bytes32 _s
+  ) external override {}
+
+  function setNFTDescriptor(IDCAHubPositionDescriptor _descriptor) external override {}
```
nchamo, line 60: Same `TODO` here

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.7 <0.9.0;
+
+import '@openzeppelin/contracts/token/ERC721/extensions/ERC721Enumerable.sol';
+import '../../interfaces/IDCAStrategies.sol';
+
+abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHandler, ERC721Enumerable {
```
nchamo, line 7: Using `ERC721Enumerable` is very expensive in terms of gas costs. But we may need to have a 
`function totalSupply() external view returns (uint256)` exposed like [here](https://github.com/Mean-Finance/dca-v2-core/blob/main/contracts/interfaces/IDCAPermissionManager.sol#L7)

```suggestion
abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHandler, ERC721 {
```

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -0,0 +1,249 @@
...
+    address owner;
+    string name;
+    ShareOfToken[] tokens;
+  }
+}
+
+interface IDCAStrategiesManagementHandler is IDCAStrategiesBase {
+  function getStrategy(uint80 _strategyId) external view returns (Strategy memory);
+
+  function getStrategyIdByName(string memory _strategyName) external view returns (uint80 _strategyId);
```
nchamo, line 43: Could you remove the underscore from these and other functions in the interface?
We usually go with the following approach:
- Interfaces have no underscore since they are only describing the parameters
- Implementations have the underscore since they might clash with other functions/variables
```suggestion
  function getStrategy(uint80 strategyId) external view returns (Strategy memory);

  function getStrategyIdByName(string memory strategyName) external view returns (uint80 strategyId);
```

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.7 <0.9.0;
+
+import '../../interfaces/IDCAStrategies.sol';
+
+abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
+  function getStrategy(uint80 _strategyId) external view override returns (Strategy memory) {}
```
nchamo, line 7: Since you already created the interfaces, could you add this to all functions in all handlers?
```suggestion
  /// @inheritdoc IDCAStrategiesManagementHandler
  function getStrategy(uint80 _strategyId) external view override returns (Strategy memory) {}
```

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.7 <0.9.0;
+
+import '../../interfaces/IDCAStrategies.sol';
+
+abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandler {
+  function deposit(
+    uint80 _strategyId,
```
nchamo, line 8: We will probably need to take the hub's address here
```suggestion
    IDCAHub _hub,
    uint80 _strategyId,
```

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -0,0 +1,249 @@
...
+  enum Permission {
+    INCREASE,
+    REDUCE,
+    WITHDRAW,
+    TERMINATE,
+    SYNC
+  }
+
+  struct Position {
+    address hub; // 20 bytes
```
nchamo, line 17:
```suggestion
    IDCAHub hub; // 20 bytes
```

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.7 <0.9.0;
+
+import '../../interfaces/IDCAStrategies.sol';
+
+abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandler {
```
nchamo, line 6: Could we add a `TODO` to add something similar to [this](https://github.com/Mean-Finance/dca-v2-core/blob/main/contracts/interfaces/IDCAHub.sol#L243)?

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -0,0 +1,52 @@
...
+  function deposit(
+    uint80 _strategyId,
+    address _from,
+    uint256 _amount,
+    uint256 _amountOfSwaps,
+    uint256 _swapInterval,
```
nchamo, line 12: These two should be `uint32`. Could you change it in other functions also?

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -0,0 +1,249 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.7 <0.9.0;
+
+import '@openzeppelin/contracts/token/ERC721/extensions/IERC721Enumerable.sol';
+import '@mean-finance/nft-descriptors/solidity/interfaces/IDCAHubPositionDescriptor.sol';
+
+interface IDCAStrategiesBase {
```
nchamo, line 7: We normally have all shared structs/enum in the main interface (in this case `IDCAStrategies`), but I'd like to hear your thoughts. Do you think it would be better to have it like this?

nchamo (commented):
Amazing work @TheAustrian1998 !

Added a few comments, but I love where this is heading

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
TheAustrian1998: I will skip this for now because I'm getting **_stack too deep_** on that function 💀. Maybe we can store the hub in _Strategy_ struct?

nchamo: Damn you solidity!!!

In that case, let's do like Sam [here](https://github.com/Mean-Finance/dca-v2-periphery/blob/main/contracts/interfaces/IDCAHubSwapper.sol#L151), and take a struct as a parameter

contracts/interfaces/IDCAStrategies.sol
TheAustrian1998: If I move shared structs then the handlers cannot find them. Do you have an example of how I need to do it?

nchamo: You would need to write `IDCAStrategies.{The other part}`

Like here: https://github.com/Mean-Finance/dca-v2-core/blob/main/contracts/DCAHub/DCAHubPositionHandler.sol#L62

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
TheAustrian1998: That's better

contracts/interfaces/IDCAStrategies.sol
TheAustrian1998: I didn't know that. Done!

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -0,0 +1,235 @@
...
+interface IDCAStrategies is IDCAStrategiesManagementHandler, IDCAStrategiesPermissionsHandler, IDCAStrategiesPositionsHandler {
+  enum Permission {
+    INCREASE,
+    REDUCE,
+    WITHDRAW,
+    TERMINATE,
+    SYNC
+  }
+
+  struct DepositParams {
```
nchamo, line 202: I think this one can be moved under `IDCAStrategiesPositionsHandler`, I don't think it will be used by others

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -0,0 +1,235 @@
...
+    uint80 strategyId;
+    address from;
+    uint256 amount;
+    uint32 amountOfSwaps;
+    uint32 swapInterval;
+    address owner;
+    PermissionSet[] permissions;
+  }
+
+  struct Position {
```
nchamo, line 213: I think this one can be moved under `IDCAStrategiesPositionsHandler`, I don't think it will be used by others

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -0,0 +1,235 @@
...
+    address token;
+    uint80 share; // 0 < share < 100%
+  }
+
+  struct PermissionSet {
+    address operator;
+    Permission[] permissions;
+  }
+
+  struct Strategy {
```
nchamo, line 230: I think this one can be moved under `IDCAStrategiesManagementHandler`, I don't think it will be used by others

nchamo (approved):
Amazing work @TheAustrian1998 🎉

PR #133, opened 2022-08-20
test: add swap multi calls

0xsambugs:
We are now adding tests that perform swaps. Such as:
- Swap & deposit
- Swap & increase
- Reduce & swap
- Withdraw & swap
- Terminate & swap

### DIFF
--- a/test/integration/DCAHubCompanion/multi-call.spec.ts
+++ b/test/integration/DCAHubCompanion/multi-call.spec.ts
@@ -11,12 +11,12 @@ import { SwapperRegistry } from '@mean-finance/swappers/typechained';
 import { TransformerOracle } from '@mean-finance/oracles/typechained';
 import { abi as DCA_HUB_ABI } from '@mean-finance/dca-v2-core/artifacts/contracts/DCAHub/DCAHub.sol/DCAHub.json';
 import { abi as IERC20_ABI } from '@openzeppelin/contracts/build/contracts/IERC20.json';
-import { BigNumber, BigNumberish, utils, Wallet } from 'ethers';
+import { BigNumber, BigNumberish, BytesLike, utils, Wallet } from 'ethers';
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers';
 import { SwapInterval } from '@test-utils/interval-utils';
-import forkBlockNumber from '@integration/fork-block-numbers';
 import { fromRpcSig } from 'ethereumjs-util';
 import { deploy } from '@integration/utils';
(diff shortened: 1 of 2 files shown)


test/integration/DCAHubCompanion/multi-call.spec.ts
```diff
@@ -87,15 +85,210 @@ contract('Multicall', () => {
...
+            const { positionId: createdPositionId, unswappedBalance } = await depositAndSwap({ from: WETH, to: USDC, amount: ETH_1 });
+            const takeData = await takeFromCallerDataIfUSDC({ from, amount: USDC_1000 });
+            const { swapExecutionData, expectedAmountOut } = await runSwapData({ tokenIn: await getAddress(from), amountIn: AMOUNT_IN, swap });
+            const permissionData = await givePermissionToCompanionData({
+              signer: positionOwner,
+              positionId: createdPositionId,
+              permissions: [Permission.INCREASE],
+            });
+            const increaseData = await increaseAllInCompanionData({ positionId: createdPositionId });
+            await DCAHubCompanion.multicall(filterMulticalls([takeData, swapExecutionData, permissionData, increaseData]), {
```
0xged, line 145: I don't get what `filterMulticalls` is intended for.

test/integration/DCAHubCompanion/multi-call.spec.ts
```diff
@@ -87,15 +85,210 @@ contract('Multicall', () => {
...
+          expectedAmountOutBTC = _expectedAmountOutBTC;
+        });
+        then('recipient has expected amount of wBTC balance', async () => {
+          const balance = await WBTC.balanceOf(recipient.address);
+          expect(balance.gte(expectedAmountOutBTC)).to.be.true;
+        });
+        then('recipient has expected amount of ETH balance', async () => {
+          const balance = await ethers.provider.getBalance(recipient.address);
+          expect(balance.gte(recipientInitialETHBalance.add(expectedAmountOutETH))).to.be.true;
+        });
```
0xged, line 272: Not checking position was terminated, but we can induce it from all the other info tho so I'm cool with that.

0xged (approved):
😍  amazing work!

test/integration/DCAHubCompanion/multi-call.spec.ts
0xsambugs: `takeData` might be `undefined`, since we don't need to take anything in the case of `from == ETH`. So `filterMulticalls` just removed said `undefined` from the array

PR #134, opened 2022-08-22
feat: permissions-functions

TheAustrian1998:
- add governable
- add permission math library
- add internal functions
- add code to external functions

I need help on [this](https://github.com/Mean-Finance/dca-v2-periphery/blob/feat/permissions-functions/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol#L57) and [this](https://github.com/Mean-Finance/dca-v2-periphery/blob/feat/permissions-functions/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol#L160), also I'm not sure about the name of [this function](https://github.com/Mean-Finance/dca-v2-periphery/blob/feat/permissions-functions/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol#L173). I tought about rename __mint_ and __burn_ to __create_ and __destroy_

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
@@ -4,6 +4,7 @@ pragma solidity >=0.8.7 <0.9.0;
 import '@openzeppelin/contracts/token/ERC721/ERC721.sol';
 import '@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol';
 import '../../interfaces/IDCAStrategies.sol';
+import '../../libraries/PermissionMath.sol';
 import '../../utils/Governable.sol';
 
 abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHandler, ERC721, EIP712 {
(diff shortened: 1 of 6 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
```diff
@@ -2,51 +2,105 @@
...
+  /// @inheritdoc IDCAStrategiesPermissionsHandler
+  bytes32 public constant PERMISSION_SET_TYPEHASH = keccak256('PermissionSet(address operator,uint8[] permissions)');
   /// @inheritdoc IDCAStrategiesPermissionsHandler
   // TODO: update this after building the new descriptor
-  function nftDescriptor() external override returns (IDCAHubPositionDescriptor) {}
+  IDCAHubPositionDescriptor public nftDescriptor;
+  /// @inheritdoc IDCAStrategiesPermissionsHandler
+  mapping(address => uint256) public nonces;
+  mapping(uint256 => uint256) public lastOwnershipChange;
+  mapping(uint256 => mapping(address => TokenPermission)) public tokenPermissions;
```
nchamo, line 36: I think it would be better if instead of using a nest mapping, we did something like [this](https://github.com/Mean-Finance/dca-v2-periphery/blob/main/contracts/DCAFeeManager/DCAFeeManager.sol#L21)

We are basically hashing a two part key into one, so we don't need to read storage twice

nchamo (commented):
I'm leaving some comments, but let's try to make each change more granular

PR #135, opened 2022-08-22
feat: token permit

0xsambugs:
We are now adding a way to execute a permit on a ERC20 token, so that we can permit + deposit with a multicall


0xged (approved):
Straight up fire. Great idea.

PR #138, opened 2022-08-23
feat: set permissions internal

TheAustrian1998:
- add code for _setPermissions and related vars
- add code for _hasPermission and related vars
- tests everything

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
@@ -2,6 +2,7 @@
 pragma solidity >=0.8.7 <0.9.0;
 
 import '@openzeppelin/contracts/token/ERC721/ERC721.sol';
+import '../../libraries/PermissionMath.sol';
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHandler, ERC721 {
@@ -12,7 +13,11 @@ abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHa
     uint248 lastUpdated;
   }
 
-  // mapping(bytes32 => TokenPermission) public tokenPermissions; // key(id, operator) => TokenPermission
+  using PermissionMath for IDCAStrategies.Permission[];
+  using PermissionMath for uint8;
+
+  mapping(uint256 => uint256) public lastOwnershipChange;
+  mapping(bytes32 => TokenPermission) public tokenPermissions; // key(id, operator) => TokenPermission
   uint256 internal _burnCounter;
   uint256 internal _mintCounter;
 
@@ -47,7 +52,14 @@ abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHa
     uint256 _id,
     address _account,
     IDCAStrategies.Permission _permission
-  ) external view override returns (bool) {}
+  ) external view override returns (bool) {
+    if (ownerOf(_id) == _account) {
+      return true;
+    }
+    TokenPermission memory _tokenPermission = tokenPermissions[keccak256(abi.encodePacked(_id, _account))];
+    // If there was an ownership change after the permission was last updated, then the address doesn't have the permission
+    return _tokenPermission.permissions.hasPermission(_permission) && lastOwnershipChange[_id] < _tokenPermission.lastUpdated;
+  }
 
   /// @inheritdoc IDCAStrategiesPermissionsHandler
   function hasPermissions(
@@ -95,18 +107,18 @@ abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHa
   }
 
   function _setPermissions(uint256 _id, IDCAStrategies.PermissionSet[] calldata _permissions) internal {
-    // uint248 _blockNumber = uint248(_getBlockNumber());
-    // for (uint256 i; i < _permissions.length; i++) {
-    //   if (_permissions[i].permissions.length == 0) {
-    //     delete tokenPermissions[keccak256(abi.encodePacked(_id, _permissions[i].operator))];
-    //   } else {
-    //     // TODO: remove this comment when adding permission math library
-    //     tokenPermissions[keccak256(abi.encodePacked(_id, _permissions[i].operator))] = TokenPermission({
-    //       permissions: _permissions[i].permissions.toUInt8(),
-    //       lastUpdated: _blockNumber
-    //     });
-    //   }
-    // }
+    uint248 _blockNumber = uint248(_getBlockNumber());
+    for (uint256 i; i < _permissions.length; i++) {
+      if (_permissions[i].permissions.length == 0) {
+        delete tokenPermissions[keccak256(abi.encodePacked(_id, _permissions[i].operator))];
+      } else {
+        // TODO: remove this comment when adding permission math library
+        tokenPermissions[keccak256(abi.encodePacked(_id, _permissions[i].operator))] = TokenPermission({
+          permissions: _permissions[i].permissions.toUInt8(),
+          lastUpdated: _blockNumber
+        });
+      }
+    }
   }
 
   // Note: virtual so that it can be overriden in tests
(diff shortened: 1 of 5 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
```diff
@@ -47,7 +52,14 @@ abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHa
     uint256 _id,
     address _account,
     IDCAStrategies.Permission _permission
-  ) external view override returns (bool) {}
+  ) external view override returns (bool) {
+    if (ownerOf(_id) == _account) {
+      return true;
+    }
+    TokenPermission memory _tokenPermission = tokenPermissions[keccak256(abi.encodePacked(_id, _account))];
```
nchamo, line 59: Here, you should be able to use the new public function I mentioned above

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
```diff
@@ -95,18 +107,17 @@ abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHa
...
-    //   } else {
-    //     // TODO: remove this comment when adding permission math library
-    //     tokenPermissions[keccak256(abi.encodePacked(_id, _permissions[i].operator))] = TokenPermission({
-    //       permissions: _permissions[i].permissions.toUInt8(),
-    //       lastUpdated: _blockNumber
-    //     });
-    //   }
-    // }
+    uint248 _blockNumber = uint248(_getBlockNumber());
+    for (uint256 i; i < _permissions.length; i++) {
```
nchamo, line 111: Look at what Sam did in `_setPermissions` [here](https://github.com/Mean-Finance/dca-v2-core/pull/232/files). By copying the struct to memory, you can avoid accessing the array multiple times, making it cheaper

contracts/mocks/DCAStrategies/DCAStrategiesPermissionsHandlerMock.sol
```diff
@@ -16,6 +16,10 @@ contract DCAStrategiesPermissionsHandlerMock is DCAStrategiesPermissionsHandler
     }
   }
 
+  function getBlockNumber() external view returns (uint256) {
```
nchamo, line 19: With my comment below, I think you can avoid this

test/unit/DCAStrategies/dca-strategies-permissions-handler.spec.ts
```diff
@@ -103,6 +104,52 @@ contract('DCAStrategiesPermissionsHandler', () => {
...
+    const OPERATOR = constants.NOT_ZERO_ADDRESS;
+    let owner: Wallet;
+    let blockNumber: BigNumber;
+    given(async () => {
+      owner = await wallet.generateRandom();
+      await DCAStrategiesPermissionsHandlerMock.mint(owner.address, [{ operator: OPERATOR, permissions: [] }]);
+      await DCAStrategiesPermissionsHandlerMock.setPermissions(tokenId, [
+        { operator: OPERATOR, permissions: [Permission.INCREASE, Permission.REDUCE] },
+      ]);
+      blockNumber = await DCAStrategiesPermissionsHandlerMock.getBlockNumber();
```
nchamo, line 118: Hardhat can be kind of annoying with this, in the sense that between the `setPermissions` call and `getBlockNumber`, the block number could change. I think it would be better if you took a random block number, and set it. That way, you can make sure that it works as expected

test/unit/DCAStrategies/dca-strategies-permissions-handler.spec.ts
```diff
@@ -103,6 +104,52 @@ contract('DCAStrategiesPermissionsHandler', () => {
     });
   });
 
+  describe('set permissions', () => {
```
nchamo, line 107: You are already testing this in `mint`. I think you can do one of two things here:
1. Test this is `mint` only
2. Test this here only, and in `mint` check that `setPermissions` was called correctly

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
```diff
@@ -12,7 +13,11 @@ abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHa
     uint248 lastUpdated;
   }
 
-  // mapping(bytes32 => TokenPermission) public tokenPermissions; // key(id, operator) => TokenPermission
+  using PermissionMath for IDCAStrategies.Permission[];
+  using PermissionMath for uint8;
+
+  mapping(uint256 => uint256) public lastOwnershipChange;
+  mapping(bytes32 => TokenPermission) public tokenPermissions; // key(id, operator) => TokenPermission
```
nchamo, line 20: Could we make this internal and add a public function that takes the id + account and returns the `TokenPermission`? So it's easier to use from outside

nchamo (commented):
Amazing work! Love the granularity ❤️

test/unit/DCAStrategies/dca-strategies-permissions-handler.spec.ts
TheAustrian1998: I will go with option 2

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol
```diff
@@ -95,18 +100,28 @@ abstract contract DCAStrategiesPermissionsHandler is IDCAStrategiesPermissionsHa
...
+    }
+  }
+
+  function _getPermissionKey(uint256 _id, address _operator) internal pure returns (bytes32) {
+    return keccak256(abi.encodePacked(_id, _operator));
+  }
+
+  function getTokenPermissions(uint256 _id, address _operator) public view override returns (TokenPermission memory) {
+    TokenPermission memory _tokenPermission = _tokenPermissions[_getPermissionKey(_id, _operator)];
+    return _tokenPermission;
```
nchamo, line 124:
```suggestion
    return _tokenPermissions[_getPermissionKey(_id, _operator)];
```

test/unit/DCAStrategies/dca-strategies-permissions-handler.spec.ts
```diff
@@ -77,22 +77,9 @@ contract('DCAStrategiesPermissionsHandler', () => {
...
-      //   then('permissions are assigned properly', async () => {
-      //     expect(await DCAStrategiesPermissionsHandlerMock.hasPermission(TOKEN_ID, OPERATOR, Permission.WITHDRAW)).to.be.true;
-      //   });
-
-      //   then('no extra permissions are assigned', async () => {
-      //     for (const permission of [Permission.INCREASE, Permission.REDUCE, Permission.TERMINATE]) {
-      //       expect(await DCAStrategiesPermissionsHandlerMock.hasPermission(TOKEN_ID, OPERATOR, permission)).to.be.false;
-      //     }
-      //   });
+      then('permissions are assigned properly', async () => {
```
nchamo, line 80: Since you prefer to go with option 2, I recommend you do something like [this](https://github.com/Mean-Finance/dca-v2-core/blob/main/contracts/mocks/DCAPermissionsManager/DCAPermissionsManager.sol#L22)
1. In the mock you override the actuall call to `_setPermissions`
2. You register all calls that went through it
3. You then expose a way to get all calls
4. In `mint` you make sure that it was called once, and with the correct parameters

nchamo (commented):
Almost there!

contracts/mocks/DCAStrategies/DCAStrategiesPermissionsHandlerMock.sol
```diff
@@ -5,6 +5,7 @@ import '../../DCAStrategies/DCAStrategies/DCAStrategiesPermissionsHandler.sol';
 
 contract DCAStrategiesPermissionsHandlerMock is DCAStrategiesPermissionsHandler {
   uint256 private _blockNumber;
+  mapping(uint256 => IDCAStrategies.Permission[]) private _setPermissionsCalls;
```
nchamo, line 8: By doing this, you are only recording once per token id. It could happen that you are calling the same token id multiple times and you wouldn't realize

I suggest you create a struct of {tokenId, IDCAStrategies.Permission[]} like [this](https://github.com/Mean-Finance/dca-v2-core/blob/main/contracts/mocks/DCAPermissionsManager/DCAPermissionsManager.sol#L8)

TheAustrian1998: I had problems with doing it that way but I will find a workaround

nchamo (approved):
Amazing work!

PR #141, opened 2022-08-24
feat: permit-function

TheAustrian1998:
- add code for permit()
- add related vars
- add tests

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategies.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategies.sol
@@ -7,5 +7,5 @@ import './DCAStrategiesPositionsHandler.sol';
 import '../../interfaces/IDCAStrategies.sol';
 
 contract DCAStrategies is DCAStrategiesManagementHandler, DCAStrategiesPermissionsHandler, DCAStrategiesPositionsHandler, IDCAStrategies {
-  constructor() ERC721('Mean Finance - DCA Strategy Position', 'MF-DCA-STRAT-P') {}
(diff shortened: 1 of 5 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategies.sol
```diff
@@ -7,5 +7,5 @@ import './DCAStrategiesPositionsHandler.sol';
 import '../../interfaces/IDCAStrategies.sol';
 
 contract DCAStrategies is DCAStrategiesManagementHandler, DCAStrategiesPermissionsHandler, DCAStrategiesPositionsHandler, IDCAStrategies {
-  constructor() ERC721('Mean Finance - DCA Strategy Position', 'MF-DCA-STRAT-P') {}
+  constructor() ERC721('Mean Finance - DCA Strategy Position', 'MF-DCA-STRAT-P') EIP712('Mean Finance - DCA Strategy Position', '1') {}
```
nchamo, line 10: Maybe we chould pass name and symbol to `DCAStrategiesPermissionsHandler` and configure `EIP712` here? Or do you prefer to have it here?

nchamo (approved):
Great work! You are on 🔥

contracts/DCAStrategies/DCAStrategies/DCAStrategies.sol
TheAustrian1998: For me, It looks more clear in this way, because you see all constructor data in the same line

nchamo: Sounds good, let's leave it here then 👍

PR #145, opened 2022-08-24
feat: set-nft-descriptor-and-governable

TheAustrian1998:
- add code for setNftDescriptor()
- add governable.sol
- tests
- also, I changed some code in _behaviours.ts_ file in _utils_, because I used a new governable version


nchamo (approved):
Looks great!

PR #147, opened 2022-08-25
feat: create-strategy

TheAustrian1998:
I hope is not too much code 😅

- add code for createStrategy()
- add related vars
- new mock
- tests for createStrategy

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -10,11 +10,11 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
   }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
-  mapping(string => uint80) public strategyIdByName;
+  mapping(bytes32 => uint80) public strategyIdByName;
   /// @inheritdoc IDCAStrategiesManagementHandler
   uint80 public strategyCounter;
   mapping(uint80 => StrategyOwnerAndVersion) internal _strategies;
-  mapping(uint80 => string) internal _strategyNameById;
+  mapping(uint80 => bytes32) internal _strategyNameById;
   mapping(bytes32 => IDCAStrategies.ShareOfToken[]) internal _tokenShares;
 
   /// @inheritdoc IDCAStrategiesManagementHandler
@@ -26,13 +26,12 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
 
   /// @inheritdoc IDCAStrategiesManagementHandler
   function createStrategy(
-    string memory _strategyName,
+    bytes32 _strategyName,
     IDCAStrategies.ShareOfToken[] memory _tokens,
     address _owner
   ) external override returns (uint80 _strategyId) {
     if (_owner == address(0)) revert IDCAStrategies.ZeroAddress();
     if (strategyIdByName[_strategyName] != 0) revert NameAlreadyExists();
-    if (bytes(_strategyName).length > 32) revert NameTooLong();
     if (_checkTokenSharesSanity(_tokens) == false) revert InvalidTokenShares();
 
     _strategyId = ++strategyCounter;
@@ -55,7 +54,7 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
   function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external override {}
 
   /// @inheritdoc IDCAStrategiesManagementHandler
-  function updateStrategyName(uint80 _strategyId, string memory _newStrategyName) external override {}
+  function updateStrategyName(uint80 _strategyId, bytes32 _newStrategyName) external override {}
 
   /// @inheritdoc IDCAStrategiesManagementHandler
   function transferStrategyOwnership(uint80 _strategyId, address _newOwner) external override {}
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -18,7 +18,7 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
   mapping(bytes32 => IDCAStrategies.ShareOfToken[]) internal _tokenShares;
 
   /// @inheritdoc IDCAStrategiesManagementHandler
-  function getStrategy(uint80 _strategyId) external view override returns (Strategy memory) {
+  function getStrategy(uint80 _strategyId) external view returns (Strategy memory) {
     StrategyOwnerAndVersion memory _strategy = _strategies[_strategyId];
     IDCAStrategies.ShareOfToken[] memory _tokens = _tokenShares[_getStrategyAndVersionKey(_strategyId, _strategy.latestVersion)];
     return Strategy({owner: _strategy.owner, name: _strategyNameById[_strategyId], currentVersion: _strategy.latestVersion, tokens: _tokens});
@@ -29,7 +29,7 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
     bytes32 _strategyName,
     IDCAStrategies.ShareOfToken[] memory _tokens,
     address _owner
-  ) external override returns (uint80 _strategyId) {
+  ) external returns (uint80 _strategyId) {
     if (_owner == address(0)) revert IDCAStrategies.ZeroAddress();
     if (strategyIdByName[_strategyName] != 0) revert NameAlreadyExists();
     if (_checkTokenSharesSanity(_tokens) == false) revert InvalidTokenShares();
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -9,45 +9,56 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
     uint96 latestVersion;
   }
 
-  mapping(uint80 => StrategyOwnerAndVersion) internal _strategies;
-  mapping(string => uint80) public getStrategyIdByName;
-  mapping(uint80 => string) public getStrategyNameById;
-  mapping(bytes32 => IDCAStrategies.ShareOfToken) internal _tokenShares;
+  /// @inheritdoc IDCAStrategiesManagementHandler
+  mapping(string => uint80) public strategyIdByName;
+  /// @inheritdoc IDCAStrategiesManagementHandler
   uint80 public strategyCounter;
+  mapping(uint80 => StrategyOwnerAndVersion) internal _strategies;
+  mapping(uint80 => string) internal _strategyNameById;
+  mapping(bytes32 => IDCAStrategies.ShareOfToken[]) internal _tokenShares;
 
-  function _checkTokenSharesSanity(IDCAStrategies.ShareOfToken memory _tokens) internal pure returns (bool) {
+  function _checkTokenSharesSanity(IDCAStrategies.ShareOfToken[] memory _tokens) internal pure returns (bool) {
     // TODO
     return true;
   }
 
-  function _getStrategyAndVersionKey(uint256 _strategyId, uint80 _version) internal pure returns (bytes32) {
+  function _getStrategyAndVersionKey(uint256 _strategyId, uint96 _version) internal pure returns (bytes32) {
     return keccak256(abi.encodePacked(_strategyId, _version));
   }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
   function getStrategy(uint80 _strategyId) external view override returns (Strategy memory) {
     StrategyOwnerAndVersion memory _strategy = _strategies[_strategyId];
-    return Strategy({owner: _strategy.owner, name: getStrategyNameById[_strategyId], currentVersion: _strategy.latestVersion});
+    IDCAStrategies.ShareOfToken[] memory _tokens = _tokenShares[_getStrategyAndVersionKey(_strategyId, _strategy.latestVersion)];
+    return Strategy({owner: _strategy.owner, name: _strategyNameById[_strategyId], currentVersion: _strategy.latestVersion, tokens: _tokens});
   }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
   function createStrategy(
     string memory _strategyName,
-    IDCAStrategies.ShareOfToken memory _tokens,
+    IDCAStrategies.ShareOfToken[] memory _tokens,
     address _owner
   ) external override returns (uint80 _strategyId) {
     if (_owner == address(0)) revert IDCAStrategies.ZeroAddress();
-    if (getStrategyIdByName[_strategyName] != 0) revert NameAlreadyExists();
-    if (bytes(_strategyName).length < 32) revert NameTooLong();
-    if (_checkTokenSharesSanity(_tokens) == false) revert BadTokenShares();
+    if (strategyIdByName[_strategyName] != 0) revert NameAlreadyExists();
+    if (bytes(_strategyName).length > 32) revert NameTooLong();
+    if (_checkTokenSharesSanity(_tokens) == false) revert InvalidTokenShares();
 
     _strategyId = ++strategyCounter;
+    bytes32 _key = _getStrategyAndVersionKey(_strategyId, 1);
+    for (uint256 i = 0; i < _tokens.length; ) {
+      _tokenShares[_key].push(_tokens[i]);
+      unchecked {
+        i++;
+      }
+    }
+
     StrategyOwnerAndVersion memory _newStrategy = StrategyOwnerAndVersion({owner: _owner, latestVersion: 1});
     _strategies[_strategyId] = _newStrategy;
-    _tokenShares[_getStrategyAndVersionKey(_strategyId, 1)] = _tokens;
-    getStrategyIdByName[_strategyName] = _strategyId;
+    strategyIdByName[_strategyName] = _strategyId;
+    _strategyNameById[_strategyId] = _strategyName;
 
-    // emit StrategyCreated(_strategyId, _newStrategy, _tokens);
+    emit StrategyCreated(_strategyId, _strategyName, _tokens, _owner);
   }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -4,18 +4,45 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
-  /// @inheritdoc IDCAStrategiesManagementHandler
-  function getStrategy(uint80 _strategyId) external view override returns (IDCAStrategies.Strategy memory) {}
+  mapping(uint80 => Strategy) internal _strategies;
+  mapping(string => uint80) internal _strategyNames;
+  mapping(bytes32 => IDCAStrategies.ShareOfToken) internal _tokenShares;
+  uint80 public strategyCounter; // NOTE: could we use _mintCounter from DCAStrategiesPermissionsHandler?
+
+  function _checkTokenSharesSanity(IDCAStrategies.ShareOfToken memory _tokens) internal pure returns (bool) {
+    // TODO
+    return true;
+  }
+
+  function _getTokenSharesKey(uint256 _strategyId, uint80 _version) internal pure returns (bytes32) {
+    // NOTE:
+    return keccak256(abi.encodePacked(_strategyId, _version));
+  }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
-  function getStrategyIdByName(string memory _strategyName) external view override returns (uint80 _strategyId) {}
+  function getStrategy(uint80 _strategyId) external view virtual override returns (IDCAStrategies.Strategy memory) {}
+
+  /// @inheritdoc IDCAStrategiesManagementHandler
+  function getStrategyIdByName(string memory _strategyName) external view virtual override returns (uint80 _strategyId) {}
 
   /// @inheritdoc IDCAStrategiesManagementHandler
   function createStrategy(
     string memory _strategyName,
-    IDCAStrategies.ShareOfToken[] memory _tokens,
+    IDCAStrategies.ShareOfToken memory _tokens,
     address _owner
-  ) external override returns (uint80 _strategyId) {}
+  ) external override returns (uint80 _strategyId) {
+    if (_owner == address(0)) revert IDCAStrategies.ZeroAddress(); // NOTE: should users be able to assign ownership to any address? what if they use it for scamming? e.g: give ownership of a strategy to vitalik.eth and use it to trick users into buying shitcoins
+    if (_strategyNames[_strategyName] != 0) revert NameAlreadyExists();
+    if (_checkTokenSharesSanity(_tokens) == false) revert BadTokenShares();
+
+    _strategyId = ++strategyCounter;
+    Strategy memory _newStrategy = Strategy({owner: _owner, name: _strategyName, version: 1});
+    _strategies[_strategyId] = _newStrategy;
+    _tokenShares[_getTokenSharesKey(_strategyId, 1)] = _tokens;
+    _strategyNames[_strategyName] = _strategyId;
+
+    emit StrategyCreated(_strategyId, _newStrategy, _tokens);
+  }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
   function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external override {}
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -6,7 +6,7 @@ import '../../interfaces/IDCAStrategies.sol';
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
   struct StrategyOwnerAndVersion {
     address owner;
-    uint96 latestVersion;
+    uint16 latestVersion;
   }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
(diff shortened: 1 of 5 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,45 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
+  mapping(uint80 => Strategy) internal _strategies;
+  mapping(string => uint80) internal _strategyNames;
+  mapping(bytes32 => IDCAStrategies.ShareOfToken) internal _tokenShares;
+  uint80 public strategyCounter; // NOTE: could we use _mintCounter from DCAStrategiesPermissionsHandler?
```
nchamo, line 10: They are different things. `_mintCounter` will count positions, while this counter will count strategies. So it's ok that we have this

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,45 @@ pragma solidity >=0.8.7 <0.9.0;
...
+
+  function _checkTokenSharesSanity(IDCAStrategies.ShareOfToken memory _tokens) internal pure returns (bool) {
+    // TODO
+    return true;
+  }
+
+  function _getTokenSharesKey(uint256 _strategyId, uint80 _version) internal pure returns (bytes32) {
+    // NOTE:
```
nchamo, line 18: Now I'm curious haha

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,45 @@ pragma solidity >=0.8.7 <0.9.0;
...
 
   /// @inheritdoc IDCAStrategiesManagementHandler
   function createStrategy(
     string memory _strategyName,
-    IDCAStrategies.ShareOfToken[] memory _tokens,
+    IDCAStrategies.ShareOfToken memory _tokens,
     address _owner
-  ) external override returns (uint80 _strategyId) {}
+  ) external override returns (uint80 _strategyId) {
+    if (_owner == address(0)) revert IDCAStrategies.ZeroAddress(); // NOTE: should users be able to assign ownership to any address? what if they use it for scamming? e.g: give ownership of a strategy to vitalik.eth and use it to trick users into buying shitcoins
```
nchamo, line 34: I see your point, but I think it's ok the way it is. It's the same with tokens, you just get them, you can't say you don't want them

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,45 @@ pragma solidity >=0.8.7 <0.9.0;
...
+  mapping(string => uint80) internal _strategyNames;
+  mapping(bytes32 => IDCAStrategies.ShareOfToken) internal _tokenShares;
+  uint80 public strategyCounter; // NOTE: could we use _mintCounter from DCAStrategiesPermissionsHandler?
+
+  function _checkTokenSharesSanity(IDCAStrategies.ShareOfToken memory _tokens) internal pure returns (bool) {
+    // TODO
+    return true;
+  }
+
+  function _getTokenSharesKey(uint256 _strategyId, uint80 _version) internal pure returns (bytes32) {
```
nchamo, line 17: Since we might use it for other things, it would make sense to name it a little bit differently
```suggestion
  function _getStrategyAndVersionKey(uint256 _strategyId, uint80 _version) internal pure returns (bytes32) {
```

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,45 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
+  mapping(uint80 => Strategy) internal _strategies;
+  mapping(string => uint80) internal _strategyNames;
```
nchamo, line 8: I think you can make this public, call it `getStrategyIdByName` and simply delete the function below. Right?

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,45 @@ pragma solidity >=0.8.7 <0.9.0;
...
   /// @inheritdoc IDCAStrategiesManagementHandler
   function createStrategy(
     string memory _strategyName,
-    IDCAStrategies.ShareOfToken[] memory _tokens,
+    IDCAStrategies.ShareOfToken memory _tokens,
     address _owner
-  ) external override returns (uint80 _strategyId) {}
+  ) external override returns (uint80 _strategyId) {
+    if (_owner == address(0)) revert IDCAStrategies.ZeroAddress(); // NOTE: should users be able to assign ownership to any address? what if they use it for scamming? e.g: give ownership of a strategy to vitalik.eth and use it to trick users into buying shitcoins
+    if (_strategyNames[_strategyName] != 0) revert NameAlreadyExists();
```
nchamo, line 35: Let's also check that the name is <= 32 bytes. I think you can do something like `bytes(_strategyName).length`

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,45 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
+  mapping(uint80 => Strategy) internal _strategies;
```
nchamo, line 7: Instead of using this struct, I think we can do something like:
**1. Create a new struct**
```
struct StrategyOwnerAndVersion {
    address owner;
    uint96 latestVersion;
}
```
Why am I suggesting this? First, Solidity handles storage in slots of 32 bytes. Addresses are 20 bytes long, so we have 12 bytes (or 96 bits) for the version. Also, we will almost never try to read the name of a Strategy on-chain. It will be used mostly for off-chain purposes, so it makes sense to split it
_Note: this struct should be internal and it should not be on the interface, since it's only used a way to store data_ 

**2. Create a new mapping**
We can create a new mapping, `mapping(uint80 => string)` for the names

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,45 @@ pragma solidity >=0.8.7 <0.9.0;
...
+  }
+
+  function _getTokenSharesKey(uint256 _strategyId, uint80 _version) internal pure returns (bytes32) {
+    // NOTE:
+    return keccak256(abi.encodePacked(_strategyId, _version));
+  }
+
   /// @inheritdoc IDCAStrategiesManagementHandler
-  function getStrategy(uint80 _strategyId) external view override returns (IDCAStrategies.Strategy memory) {}
+  function getStrategy(uint80 _strategyId) external view virtual override returns (IDCAStrategies.Strategy memory) {}
```
nchamo, line 23: I think it would be better if we implemented this function instead of making it virtual and implementing it on the tests

It makes sense to implement it if it helps with testing

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -6,10 +6,26 @@ import '@mean-finance/dca-v2-core/contracts/interfaces/IDCAHub.sol';
...
+  error NameAlreadyExists();
+
+  /// @notice Thrown when a provided array of token shares is misconfigured
+  error BadTokenShares();
+
   struct Strategy {
     address owner;
     string name;
-    IDCAStrategies.ShareOfToken[] tokens;
+    uint80 version;
```
nchamo, line 28:
```suggestion
    uint96 currentVersion;
```

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -6,10 +6,26 @@ import '@mean-finance/dca-v2-core/contracts/interfaces/IDCAHub.sol';
...
+  /// @notice Thrown when a provided strategy name already exist
+  error NameAlreadyExists();
+
+  /// @notice Thrown when a provided array of token shares is misconfigured
+  error BadTokenShares();
+
   struct Strategy {
     address owner;
     string name;
-    IDCAStrategies.ShareOfToken[] tokens;
```
nchamo, line 12: Let's not delete this. The idea for this struct is to be used for off-chain purposes, so it makes sense to have everything in one place

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -304,8 +320,8 @@ interface IDCAStrategies is IDCAStrategiesManagementHandler, IDCAStrategiesPermi
   }
 
   struct ShareOfToken {
```
nchamo, line 322: Why are you doing this change? If you have `address[] addresses, uint80[] shares`, then you need to validate that `addresses.length == shares.length`. If you have `struct[]` then you don't need to validate it

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -304,8 +320,8 @@ interface IDCAStrategies is IDCAStrategiesManagementHandler, IDCAStrategiesPermi
...
-    address token;
-    uint80 share; // 0 < share < 100%
+    address[] tokens;
+    uint80[] shares; // 0 < share < 100%
```
nchamo, line 324: Just realized that this can be `uint96` instead of `uint80`

contracts/mocks/DCAStrategies/DCAStrategiesManagementHandlerMock.sol
```diff
@@ -0,0 +1,18 @@
...
+pragma solidity >=0.8.7 <0.9.0;
+
+import '../../DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol';
+
+contract DCAStrategiesManagementHandlerMock is DCAStrategiesManagementHandler {
+  function getStrategy(uint80 _strategyId) external view override returns (Strategy memory) {
+    return _strategies[_strategyId];
+  }
+
+  function getTokenShares(uint256 _strategyId, uint80 _version) external view returns (IDCAStrategies.ShareOfToken memory) {
```
nchamo, line 11: If we implement `getStrategy`, I think we won't need this

nchamo (commented):
Off to a great start! Didn't look at the tests yet, but I'll do that after we discuss this comments

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
TheAustrian1998: You are right! I don't know how I didn't realize it

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
TheAustrian1998: I forget what I was to write there 🤔

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
TheAustrian1998: I understand, but I think we can solve what I say with:
if `owner != msg.sender` we set the `msg.sender` as the owner and `owner` parm as pending owner (like create strategy and transfer ownership but in the same tx)

nchamo: But that might be too uncomfortable for integrations. If we (or someone else) creates a contract that created strategies, then the owner will always be said contract and the actual owner will need to execute a new tx. I understand your concern, but I don't want to make this more complex to avoid that particular scenario

TheAustrian1998: Makes sense

contracts/interfaces/IDCAStrategies.sol
TheAustrian1998: It's the only way I found to bypass `memory to storage not yet supported` error

TheAustrian1998:
I will commit what I have, and you can review again. There is some things I don't understand completely, and I want to know if I did it well

contracts/interfaces/IDCAStrategies.sol
nchamo: What you need to do is:
```
function createStrategy(
    string memory _strategyName,
    IDCAStrategies.ShareOfToken[] memory _tokens,
    address _owner
  ) external override returns (uint80 _strategyId) {
    bytes32 _key = ...
    for (uint i = 0; i < _tokens.length; ) {
        _tokenShares[key].push(_tokens[i]);
        unchecked {i++;}
    }
```

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,51 @@ pragma solidity >=0.8.7 <0.9.0;
...
-  /// @inheritdoc IDCAStrategiesManagementHandler
-  function getStrategy(uint80 _strategyId) external view override returns (IDCAStrategies.Strategy memory) {}
+  struct StrategyOwnerAndVersion {
+    address owner;
+    uint96 latestVersion;
+  }
+
+  mapping(uint80 => StrategyOwnerAndVersion) internal _strategies;
+  mapping(string => uint80) public getStrategyIdByName;
+  mapping(uint80 => string) public getStrategyNameById;
```
nchamo, line 14: I think`getStrategyNameById` can be internal, since we will already expose it under `getStrategy`

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,51 @@ pragma solidity >=0.8.7 <0.9.0;
...
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
-  /// @inheritdoc IDCAStrategiesManagementHandler
-  function getStrategy(uint80 _strategyId) external view override returns (IDCAStrategies.Strategy memory) {}
+  struct StrategyOwnerAndVersion {
+    address owner;
+    uint96 latestVersion;
+  }
+
+  mapping(uint80 => StrategyOwnerAndVersion) internal _strategies;
+  mapping(string => uint80) public getStrategyIdByName;
```
nchamo, line 13: You can add the tag here
```suggestion
  /// @inheritdoc IDCAStrategiesManagementHandler
  mapping(string => uint80) public getStrategyIdByName;
```

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,51 @@ pragma solidity >=0.8.7 <0.9.0;
...
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
-  /// @inheritdoc IDCAStrategiesManagementHandler
-  function getStrategy(uint80 _strategyId) external view override returns (IDCAStrategies.Strategy memory) {}
+  struct StrategyOwnerAndVersion {
+    address owner;
+    uint96 latestVersion;
+  }
+
+  mapping(uint80 => StrategyOwnerAndVersion) internal _strategies;
```
nchamo, line 12: We normally move internal variables below the public ones

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,51 @@ pragma solidity >=0.8.7 <0.9.0;
...
+  mapping(string => uint80) public getStrategyIdByName;
+  mapping(uint80 => string) public getStrategyNameById;
+  mapping(bytes32 => IDCAStrategies.ShareOfToken) internal _tokenShares;
+  uint80 public strategyCounter;
+
+  function _checkTokenSharesSanity(IDCAStrategies.ShareOfToken memory _tokens) internal pure returns (bool) {
```
nchamo, line 18: We normaly move internal functions below the public/external ones

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
nchamo: Also, now that I see it, let's rename the variable (and the function) to `strategyIdByName`

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,51 @@ pragma solidity >=0.8.7 <0.9.0;
...
     string memory _strategyName,
-    IDCAStrategies.ShareOfToken[] memory _tokens,
+    IDCAStrategies.ShareOfToken memory _tokens,
     address _owner
-  ) external override returns (uint80 _strategyId) {}
+  ) external override returns (uint80 _strategyId) {
+    if (_owner == address(0)) revert IDCAStrategies.ZeroAddress();
+    if (getStrategyIdByName[_strategyName] != 0) revert NameAlreadyExists();
+    if (bytes(_strategyName).length < 32) revert NameTooLong();
+    if (_checkTokenSharesSanity(_tokens) == false) revert BadTokenShares();
```
nchamo, line 42:
```suggestion
    if (_checkTokenSharesSanity(_tokens) == false) revert InvalidTokenShares();
```

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -6,19 +6,40 @@ import '@mean-finance/dca-v2-core/contracts/interfaces/IDCAHub.sol';
...
+   * @param strategyId The id of the new strategy
+   * @param strategy The struct with the values of the new strategy
+   */
+  event StrategyCreated(uint80 strategyId, Strategy strategy, IDCAStrategies.ShareOfToken tokens);
+
+  /// @notice Thrown when strategy name is more than 32 bytes
+  error NameTooLong();
+
+  /// @notice Thrown when a provided array is empty
+  error LengthZero();
```
nchamo, line 20: When are you emitting this?

nchamo (commented):
Some more comments

contracts/interfaces/IDCAStrategies.sol
TheAustrian1998: It was on some piece of code I deleted, but I will use it `_checkTokenSharesSanity()`

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -9,45 +9,56 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
...
+  /// @inheritdoc IDCAStrategiesManagementHandler
+  mapping(string => uint80) public strategyIdByName;
+  /// @inheritdoc IDCAStrategiesManagementHandler
   uint80 public strategyCounter;
+  mapping(uint80 => StrategyOwnerAndVersion) internal _strategies;
+  mapping(uint80 => string) internal _strategyNameById;
+  mapping(bytes32 => IDCAStrategies.ShareOfToken[]) internal _tokenShares;
 
-  function _checkTokenSharesSanity(IDCAStrategies.ShareOfToken memory _tokens) internal pure returns (bool) {
+  function _checkTokenSharesSanity(IDCAStrategies.ShareOfToken[] memory _tokens) internal pure returns (bool) {
```
nchamo, line 20: Could you move the internal functions to the bottom of the file, after the public/external functions?

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,62 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
+  struct StrategyOwnerAndVersion {
+    address owner;
+    uint96 latestVersion;
```
nchamo, line 9: Ok, I just checked our original doc and I realized that this should be `uint16`. We will have some extra space in this struct, but it will make sense when we store positions

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,18 +4,62 @@ pragma solidity >=0.8.7 <0.9.0;
...
+    bytes32 _key = _getStrategyAndVersionKey(_strategyId, 1);
+    for (uint256 i = 0; i < _tokens.length; ) {
+      _tokenShares[_key].push(_tokens[i]);
+      unchecked {
+        i++;
+      }
+    }
+
+    StrategyOwnerAndVersion memory _newStrategy = StrategyOwnerAndVersion({owner: _owner, latestVersion: 1});
+    _strategies[_strategyId] = _newStrategy;
```
nchamo, line 57:
```suggestion
    _strategies[_strategyId] = StrategyOwnerAndVersion({owner: _owner, latestVersion: 1});
```

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -6,15 +6,39 @@ import '@mean-finance/dca-v2-core/contracts/interfaces/IDCAHub.sol';
 import '@mean-finance/nft-descriptors/solidity/interfaces/IDCAHubPositionDescriptor.sol';
 
 interface IDCAStrategiesManagementHandler {
+  /**
+   * @notice Emitted when a new strategy is created
+   * @param strategyId The id of the new strategy
+   * @param strategyName The name of the strategy
+   * @param tokens An array with all tokenShares
```
nchamo, line 13:
```suggestion
   * @param tokens An array with all token shares
```

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -6,15 +6,39 @@ import '@mean-finance/dca-v2-core/contracts/interfaces/IDCAHub.sol';
...
+  /// @notice Thrown when a provided strategy name already exist
+  error NameAlreadyExists();
+
+  /// @notice Thrown when a provided array of token shares is misconfigured
+  error InvalidTokenShares();
+
   struct Strategy {
     address owner;
     string name;
+    uint96 currentVersion;
```
nchamo, line 33:
```suggestion
    uint16 currentVersion;
```

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -295,6 +316,9 @@ interface IDCAStrategiesPositionsHandler {
 }
 
 interface IDCAStrategies is IDCAStrategiesManagementHandler, IDCAStrategiesPermissionsHandler, IDCAStrategiesPositionsHandler {
+  /// @notice Thrown when a user provides a zero address when they shouldn't
+  error ZeroAddress();
```
nchamo, line 320: Nice!

contracts/mocks/DCAStrategies/DCAStrategiesManagementHandlerMock.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.7 <0.9.0;
+
+import '../../DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol';
+
+contract DCAStrategiesManagementHandlerMock is DCAStrategiesManagementHandler {
+  function getTokenShares(uint80 _strategyId, uint96 _version) external view returns (IDCAStrategies.ShareOfToken[] memory) {
```
nchamo, line 7: Is this necessary? Can't you read it when you call `getStrategy`?

nchamo (commented):
We are almost there!

This is amazing work @TheAustrian1998

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        let _strategyId: BigNumber = await readArgFromEventOrFail(tx, 'StrategyCreated', 'strategyId');
+        let _strategyName: string = await readArgFromEventOrFail(tx, 'StrategyCreated', 'strategyName');
+        let _tokens: IDCAStrategies.ShareOfTokenStruct[] = await readArgFromEventOrFail(tx, 'StrategyCreated', 'tokens');
+        let _owner: string = await readArgFromEventOrFail(tx, 'StrategyCreated', 'owner');
+        expect(_strategyId).to.be.equal(1);
+        expect(_strategyName).to.be.equal(strategy.name);
+        expect(_owner).to.be.equal(strategy.owner);
+        _tokens.forEach((s: any, i: any) => {
+          expect(tokenShares[i].share).to.be.equal(s.share);
+          expect(tokenShares[i].token).to.be.equal(s.token);
```
nchamo, line 101:
```suggestion
          expect(strategy.tokenShares[i].share).to.be.equal(s.share);
          expect(strategy.tokenShares[i].token).to.be.equal(s.token);
```

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      then('tx reverted with message', async () => {
+        await expect(DCAStrategiesManagementHandlerMock.createStrategy(NAME, SHARES, constants.ZERO_ADDRESS)).to.be.revertedWith(
+          'ZeroAddress()'
+        );
+      });
+    });
+    when('strategy is created', () => {
+      let tx: TransactionResponse;
+      let strategy: IDCAStrategiesManagementHandler.StrategyStruct;
+      let tokenShares: IDCAStrategies.ShareOfTokenStruct[];
```
nchamo, line 59: I think you don't need this
```suggestion
```

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        expect(strategy.name).to.be.equal(NAME);
+      });
+      then('name is saved correctly in reversed mapping', async () => {
+        expect(await DCAStrategiesManagementHandlerMock.strategyIdByName(NAME)).to.be.equal(1);
+      });
+      then('version is correct', async () => {
+        expect(strategy.currentVersion).to.be.equal(1);
+      });
+      then('shares are correct', async () => {
+        SHARES.forEach((s, i) => {
```
nchamo, line 86: You should also check that `SHARES.length === strategy.tokenShares.length`. Otherwise, the test could pass but `strategy.tokenShares` could have more tokens

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      });
+      then('event is emitted', async () => {
+        let _strategyId: BigNumber = await readArgFromEventOrFail(tx, 'StrategyCreated', 'strategyId');
+        let _strategyName: string = await readArgFromEventOrFail(tx, 'StrategyCreated', 'strategyName');
+        let _tokens: IDCAStrategies.ShareOfTokenStruct[] = await readArgFromEventOrFail(tx, 'StrategyCreated', 'tokens');
+        let _owner: string = await readArgFromEventOrFail(tx, 'StrategyCreated', 'owner');
+        expect(_strategyId).to.be.equal(1);
+        expect(_strategyName).to.be.equal(strategy.name);
+        expect(_owner).to.be.equal(strategy.owner);
+        _tokens.forEach((s: any, i: any) => {
```
nchamo, line 99: Instead of using `_tokens`, why not compare against `SHARES` like the test before? Also, let's compare the length

nchamo (commented):
Code looks great, just added some comments on the tests

0xsambugs (approved):
Amazing work @TheAustrian1998 ! Truly great 🎉

PR #148, opened 2022-08-26
feat: update-strategy-tokens

TheAustrian1998:
- add code for updateStrategyTokens()
- tests

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -4,6 +4,12 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
+  modifier onlyStratOwner(uint80 _strategyId) {
+    address owner = _strategies[_strategyId].owner;
+    if (msg.sender != owner) revert OnlyStratOwner();
+    _;
+  }
+
   struct StrategyOwnerAndVersion {
     address owner;
     uint16 latestVersion;
@@ -51,7 +57,20 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
   }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
-  function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external override {}
+  function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external override onlyStratOwner(_strategyId) {
+    if (_checkTokenSharesSanity(_tokens) == false) revert InvalidTokenShares();
+
+    uint16 _newVersion = ++_strategies[_strategyId].latestVersion;
+    bytes32 _key = _getStrategyAndVersionKey(_strategyId, _newVersion);
+    for (uint256 i = 0; i < _tokens.length; ) {
+      _tokenShares[_key].push(_tokens[i]);
+      unchecked {
+        i++;
+      }
+    }
+
+    emit StrategyUpdated(_strategyId, _tokens);
+  }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
   function updateStrategyName(uint80 _strategyId, bytes32 _newStrategyName) external override {}
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -4,12 +4,6 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
-  modifier onlyStratOwner(uint80 _strategyId) {
-    address owner = _strategies[_strategyId].owner;
-    if (msg.sender != owner) revert OnlyStratOwner();
-    _;
-  }
-
   struct StrategyOwnerAndVersion {
     address owner;
     uint16 latestVersion;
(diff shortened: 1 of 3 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -51,7 +57,20 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
   }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
-  function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external override {}
+  function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external override onlyStratOwner(_strategyId) {
```
0xsambugs, line 60: I think that `override` is not longer necessary since [0.8.8](https://blog.soliditylang.org/2021/09/27/solidity-0.8.8-release-announcement/)

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,6 +4,12 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
+  modifier onlyStratOwner(uint80 _strategyId) {
```
0xsambugs, line 7: I really like the modifier, but you are reading `_strategies[_strategyId]` twice. Once here, and once to get the new version. That's one extra storage read, which we could avoid if we just do it "manually" inside `updateStrategyTokens`

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -27,6 +28,18 @@ contract('DCAStrategiesManagementHandler', () => {
     await snapshot.revert(snapshotId);
   });
 
+  function generateRandomAddress() {
```
0xsambugs, line 31: I think we already have a utils in utils/wallet that does this

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -27,6 +28,18 @@ contract('DCAStrategiesManagementHandler', () => {
...
+    return ethers.utils.getAddress(ethers.utils.hexlify(ethers.utils.randomBytes(20)));
+  }
+
+  function compareTokens(arrayA: IDCAStrategies.ShareOfTokenStruct[], arrayB: IDCAStrategies.ShareOfTokenStruct[]) {
```
0xsambugs, line 35: Makes sense to have this util function, but we usually try to leave it after the "important code". In production code, we usually do:
1. External/public
2. Internal/private

In tests, we usually do:
1. Tests
2. Utils

Does it make sense?

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -92,11 +101,55 @@ contract('DCAStrategiesManagementHandler', () => {
...
-          expect(tokens[i].token).to.be.equal(s.token);
-        });
+        compareTokens(SHARES, tokens);
+      });
+    });
+  });
+
+  describe('createStrategy', () => {
+    let owner: Wallet;
+    const NAME = ethers.utils.formatBytes32String('Optimism Ecosystem - v1');
```
0xsambugs, line 111: I think we can share this (and maybe some of the other constants) between all tests

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -92,11 +101,55 @@ contract('DCAStrategiesManagementHandler', () => {
...
+    });
+    when('sender is not the owner', () => {
+      given(async () => {
+        await DCAStrategiesManagementHandlerMock.createStrategy(NAME, SHARES_1, owner.address);
+      });
+      then('tx reverted with message', async () => {
+        await expect(DCAStrategiesManagementHandlerMock.updateStrategyTokens(1, SHARES_2)).to.be.revertedWith('OnlyStratOwner()');
+      });
+    });
+    when('strategy is updated', () => {
```
0xsambugs, line 133: Could you also test that the owner and name have not changed? Just to be on the safe side

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -51,7 +57,20 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
...
-  function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external override {}
+  function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external override onlyStratOwner(_strategyId) {
+    if (_checkTokenSharesSanity(_tokens) == false) revert InvalidTokenShares();
+
+    uint16 _newVersion = ++_strategies[_strategyId].latestVersion;
+    bytes32 _key = _getStrategyAndVersionKey(_strategyId, _newVersion);
+    for (uint256 i = 0; i < _tokens.length; ) {
+      _tokenShares[_key].push(_tokens[i]);
+      unchecked {
+        i++;
```
0xsambugs, line 68: I hate that we have to do this because Solidity compilers suck

0xsambugs (commented):
Just left a few minor comments here and there but overall looks really good! 👏👏👏👏👏

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
TheAustrian1998: Yes, but I wanted different addresses on each call. Is it mandatory that the tests be deterministic?

0xsambugs: I think it wasn't because of determinism, it was so that the gas cost didn't change each time the test was run. But this is good enough for me

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -51,7 +51,24 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
   }
 
   /// @inheritdoc IDCAStrategiesManagementHandler
-  function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external override {}
+  function updateStrategyTokens(uint80 _strategyId, IDCAStrategies.ShareOfToken[] memory _tokens) external {
+    StrategyOwnerAndVersion memory _strategy = _strategies[_strategyId];
+    if (msg.sender != _strategy.owner) revert OnlyStratOwner();
+    if (_checkTokenSharesSanity(_tokens) == false) revert InvalidTokenShares();
+
+    uint16 _newVersion = ++_strategy.latestVersion;
```
0xsambugs, line 59: We can avoid modifying the value in memory
```suggestion
    uint16 _newVersion = _strategy.latestVersion + 1;
```

0xsambugs (approved):
Thanks for the changes, great work @TheAustrian1998 !

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
TheAustrian1998: But when I save the new version number? saving the entire struct at the end of the function?

0xsambugs: You are already saving it in line 60, with `_strategies[_strategyId].latestVersion = _newVersion;`

In line 59 you are increasing the version in the memory version of the struct, which is not necessary

TheAustrian1998: Got it, I understand now

PR #150, opened 2022-08-29
feat: token-shares-limit

TheAustrian1998:
- add code for limiting token shares
- tests

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -14,9 +14,15 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
   /// @inheritdoc IDCAStrategiesManagementHandler
   uint80 public strategyCounter;
   uint16 internal constant _TOTAL = 100e2;
+  uint8 internal immutable _maxTokenShares;
   mapping(uint80 => StrategyOwnerAndVersion) internal _strategies;
   mapping(uint80 => bytes32) internal _strategyNameById;
   mapping(bytes32 => IDCAStrategies.ShareOfToken[]) internal _tokenShares;
+
+  constructor(uint8 __maxTokenShares) {
+    if (__maxTokenShares == 0) revert InvalidMaxTokenShares();
+    _maxTokenShares = __maxTokenShares;
+  }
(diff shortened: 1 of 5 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -14,10 +14,16 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
   /// @inheritdoc IDCAStrategiesManagementHandler
   uint80 public strategyCounter;
   uint16 internal constant _TOTAL = 100e2;
+  uint8 internal immutable _maxTokenShares;
```
0xsambugs, line 17: Let's make this public, so that we can read it later. Also, we use the same styling for immutables as constants, so it would look like `MAX_TOKEN_SHARES`

Finally, let's add some docs for it

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -157,6 +173,13 @@ contract('DCAStrategiesManagementHandler', () => {
         );
       });
     });
+    when('token shares exceed max amount', () => {
```
0xsambugs, line 176: Maybe something like this can help to avoid repeating the same tests:
https://github.com/Mean-Finance/dca-v2-periphery/blob/main/test/unit/DCAHubCompanion/dca-hub-companion-hub-proxy-handler.spec.ts#L224

The idea is basically to have a function like:
```
function invalidSharesTest({ method, params }: { method: string, params: (shares) => any[] }) {
    when('token shares exceed max amount', () => {
        then('tx reverted with message', async () => {
            const invalidParams = params(SHARES_EXCEED_AMOUNT)
            await expect(DCAStrategiesManagementHandlerMock.connect(user)[method](invalidParams)).to.be.revertedWith(
          'TokenSharesExceedAmount()'
            );
         });
    });
    }
}
```

And then you can call:
```
invalidSharesTest({ method: 'createStrategy', params: (shares) => [NAME, shares, user.address]})
invalidSharesTest({ method: 'updateStrategyTokens', params: (shares) => [1, shares]})

TheAustrian1998: Works perfectly!

PR #151, opened 2022-08-29
feat: transfer-strategy-ownership

TheAustrian1998:
- add code for transferStrategyOwnership()
- tests

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -8,7 +8,8 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
     address owner;
     uint16 latestVersion;
   }
-
+  /// @inheritdoc IDCAStrategiesManagementHandler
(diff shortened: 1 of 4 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -18,6 +18,7 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
   // solhint-disable-next-line var-name-mixedcase
   uint8 public immutable MAX_TOKEN_SHARES;
   mapping(uint80 => StrategyOwnerAndVersion) internal _strategies;
+  mapping(uint80 => address) internal _strategiesPendingOwner;
```
nchamo, line 21: Maybe we should make this public? We usually try to make sure that everything can be read off-chain in some form. It doesn't hurt to make it public, and if we need to figure out what is stored, it's easier than trying to read internal/private variables

Also, let's try to add a function to the interface, explaining what this does

nchamo (approved):
Amazing work @TheAustrian1998 ! Just one super minor comment

Feel free to make it public, add the description to the interface, add the `@inheritdoc` tag on the mapping, and merge

PR #158, opened 2022-09-05
feat: withdraw-swapped

TheAustrian1998:
- add code for withdrawSwapped()
- tests

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
@@ -56,23 +56,22 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
   }
 
   /// @inheritdoc IDCAStrategiesPositionsHandler
-  function withdrawSwapped(uint256 _positionId, address _recipient) external returns (TokenAmounts memory _tokenAmounts) {
-    if (!_isOwnerOf(_positionId, msg.sender)) revert IDCAStrategies.NotOwner();
+  function withdrawSwapped(uint256 _positionId, address _recipient) external returns (TokenAmounts[] memory _tokenAmounts) {
+    if (!_hasWithdrawPermission(_positionId, msg.sender)) revert NoPermissions();
 
     Position memory _position = _userPositions[_positionId];
     IDCAHub _hub = IDCAHub(_position.hub);
-    _tokenAmounts.tokens = new address[](_position.positions.length);
-    _tokenAmounts.amounts = new uint256[](_position.positions.length);
+    _tokenAmounts = new TokenAmounts[](_position.positions.length);
     for (uint256 i = 0; i < _position.positions.length; ) {
-      _tokenAmounts.amounts[i] = _hub.withdrawSwapped(_position.positions[i], _recipient);
-      _tokenAmounts.tokens[i] = address(_hub.userPosition(_position.positions[i]).to);
+      _tokenAmounts[i].amount = _hub.withdrawSwapped(_position.positions[i], _recipient);
+      _tokenAmounts[i].token = address(_hub.userPosition(_position.positions[i]).to);
 
       unchecked {
         i++;
       }
     }
 
-    emit Withdrew(msg.sender, _recipient, _positionId, _position.positions, _tokenAmounts);
+    emit Withdrew(msg.sender, _recipient, _positionId, _tokenAmounts);
   }
 
   /// @inheritdoc IDCAStrategiesPositionsHandler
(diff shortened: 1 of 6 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -56,7 +56,24 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
   }
 
   /// @inheritdoc IDCAStrategiesPositionsHandler
-  function withdrawSwapped(uint256 _positionId, address _recipient) external override returns (uint256) {}
+  function withdrawSwapped(uint256 _positionId, address _recipient) external returns (TokenAmounts memory _tokenAmounts) {
+    if (!_isOwnerOf(_positionId, msg.sender)) revert IDCAStrategies.NotOwner();
```
nchamo, line 60: You should be checking if the `msg.sender` has permissions to execute a withdraw, by calling `hasPermission`. Otherwise, the permission system doesn't make much sense

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -412,6 +425,12 @@ interface IDCAStrategiesPositionsHandler {
     IDCAStrategies.PermissionSet[] permissions;
   }
 
+  // NOTE: i'm doing it this way because of a strange solidity error
+  struct TokenAmounts {
```
nchamo, line 429: What error where you getting (we can sync offline if you want)? Because we should be returning an array of a `struct { address token; uint256 amount }`

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -397,6 +394,22 @@ interface IDCAStrategiesPositionsHandler {
...
+   * @param recipient The address of the user that will receive the withdrawn tokens
+   * @param positionId The id of the position that was affected
+   * @param underlyingsPositionId The ids of the underlying positions affected
+   * @param tokenAmounts The amounts withdrawn and respective tokens
+   */
+  event Withdrew(
+    address indexed withdrawer,
+    address indexed recipient,
+    uint256 positionId,
+    uint256[] underlyingsPositionId,
```
nchamo, line 409: I think we can avoid this. It costs gas to emit data, and this information should already be available off-chain, since we already emitted the underlying position ids in `deposit`

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -469,6 +488,9 @@ interface IDCAStrategies is IDCAStrategiesManagementHandler, IDCAStrategiesPermi
   /// @notice Thrown when a user provides a zero address when they shouldn't
   error ZeroAddress();
 
+  /// @notice Thrown when a user tries to perform an action for a token they do not own
+  error NotOwner();
```
nchamo, line 492: I think that since we are going with the permission error instead of this error, we can move this back to under `IDCAStrategiesPermissionsHandler`

nchamo (commented):
Looks pretty good 🎉 

Made some comments 😊

contracts/interfaces/IDCAStrategies.sol
TheAustrian1998: It's strange but is working right now. So... fixed? 😄

nchamo: Hahaha 🤷

nchamo: Hahaha 🤷

test/unit/DCAStrategies/dca-strategies-positions-handler.spec.ts
```diff
@@ -198,4 +202,68 @@ contract('DCAStrategiesPositionsHandler', () => {
...
+      then('event is emitted', async () => {
+        const withdrawer = await readArgFromEventOrFail(tx, 'Withdrew', 'withdrawer');
+        const recipient = await readArgFromEventOrFail(tx, 'Withdrew', 'recipient');
+        const positionId = await readArgFromEventOrFail(tx, 'Withdrew', 'positionId');
+        const tokenAmounts: IDCAStrategiesPositionsHandler.TokenAmountsStruct[] = await readArgFromEventOrFail(tx, 'Withdrew', 'tokenAmounts');
+
+        expect(withdrawer).to.be.equal(user.address);
+        expect(recipient).to.be.equal(user.address);
+        expect(positionId).to.be.equal(1);
+        tokenAmounts.forEach((ta, i) => {
```
nchamo, line 262: Let's also check that `tokenAmounts.length === tokens.length`

nchamo (commented):
Almost there, just a few minor comments

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -9,6 +9,11 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
 
   mapping(uint256 => Position) internal _userPositions;
 
+  modifier onlyWithPermission(uint256 _positionId, IDCAStrategies.Permission _permission) {
```
nchamo, line 12: Nit: since this is only for internal purposes, could you move it to the bottom of the file?

nchamo (approved):
Looks great!

PR #159, opened 2022-09-06
feat: increase-position

TheAustrian1998:
- add code for increasePosition()
- improvements
- tests

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
@@ -37,7 +37,6 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
       hub: _parameters.hub,
       strategyId: _parameters.strategyId,
       strategyVersion: _parameters.version,
-      fromToken: _parameters.from,
       positions: _positions
     });
 
@@ -81,17 +80,20 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
   /// @inheritdoc IDCAStrategiesPositionsHandler
   function increasePosition(
     uint256 _positionId,
+    address _fromToken,
     uint256 _amount,
     uint32 _newSwaps
   ) external onlyWithPermission(_positionId, IDCAStrategies.Permission.INCREASE) {
     Position memory _position = _userPositions[_positionId];
     IDCAStrategies.ShareOfToken[] memory _tokens = _getTokenShares(_position.strategyId, _position.strategyVersion);
 
-    // extract money from user
-    IERC20(_position.fromToken).safeTransferFrom(msg.sender, address(this), _amount);
+    if (_amount != 0) {
+      // extract money from user
+      IERC20(_fromToken).safeTransferFrom(msg.sender, address(this), _amount);
 
-    // approve hub (if needed)
-    _approveHub(_position.fromToken, _position.hub, _amount);
+      // approve hub (if needed)
+      _approveHub(_fromToken, _position.hub, _amount);
+    }
 
     uint16 _total = _getTotalShares();
     uint256 _amountSpent;
@@ -208,6 +210,7 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
     uint256 _total,
     uint256 _amountSpent
   ) internal pure returns (uint256 _optimal) {
+    if (_amount == 0) return 0;
     return _index < _arrayLength - 1 ? (_amount * _share) / _total : _amount - _amountSpent;
   }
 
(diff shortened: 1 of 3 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -82,7 +83,32 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
     uint256 _positionId,
     uint256 _amount,
     uint32 _newSwaps
-  ) external override {}
+  ) external onlyWithPermission(_positionId, IDCAStrategies.Permission.INCREASE) {
+    Position memory _position = _userPositions[_positionId];
+    IDCAStrategies.ShareOfToken[] memory _tokens = _getTokenShares(_position.strategyId, _position.strategyVersion);
+
+    // extract money from user
+    IERC20(_position.fromToken).safeTransferFrom(msg.sender, address(this), _amount);
```
nchamo, line 91: There is a valid case where `_amount` might be 0. This could happen when the user has some remaining funds and they want to re-distribute them in a new amounts of swaps. When that happens, it would be cheaper to avoid calling `safeTransferFrom` and `_approveHub`

contracts/interfaces/IDCAStrategies.sol
```diff
@@ -433,6 +442,7 @@ interface IDCAStrategiesPositionsHandler {
     IDCAHub hub; // 20 bytes
     uint80 strategyId; // 10 bytes
     uint16 strategyVersion; // 2 bytes
+    address fromToken;
```
nchamo, line 445: I see where you are going with this, let's explore the different options (let me know if you think that an option is missing):
1. Add this `fromToken` to the struct, so that it's stored on `deposit`
1. Pass the `from` token when calling `increase` as one of the parameters
1. Get the `from` token by querying for one of the underlying positions

**Option (1)**
Pros:
- Easy to implement

Cons:
- Adding the new field to the stuct means that it will now take two storage slots instead of 1. So it's an extra storage write + an extra storage read 

**Option (2)**
Pros:
- Easiest and cheapest to implement

Cons:
- Caller needs to know the `from` token when calling the function
- What happens if the caller sends a wrong token address? Let's explore it:
    - Let's say that the `from` token is USDC, and the caller sends the DAI address instead. If they send `0` as amount, then nothing will happen since no tokens will be taken from the caller/approved. If they send more than `0`, then `_toIncrease` will be more than zero. So when `hub.increasePosition` is called, the hub will try to take USDC tokens from the user. When that happens, the hub will revert, because this contract does not hold any USDC tokens. 
    - So we know that if the user sends the wrong address then the function will revert

**Option (3)**
Pros:
- No extra storage slots
- Nothing to pass to the function, so no change of mistake/revert

Cons:
- The contract needs to call `hub.userPosition`, which takes quite a few storage reads


Having said all this, I think I tend to go towards (2), since it's the cheapest option available, and the FE already knows the FE token. But I'm interested in what you think is best and why. Also @0xged , care to chime in?

nchamo (commented):
Added one comment that we should discuss

contracts/interfaces/IDCAStrategies.sol
TheAustrian1998: I would choose option 2, I'm just concerned by the danger of allowing the user to send an arbitrary contract in the params, (could be a potential re-entrancy because that contract is called at [this line](https://github.com/Mean-Finance/dca-v2-periphery/blob/f9cbc01ba3011950fd8cd84f77599b929c5f1574/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol#L91), right?)

nchamo: Yes, passing an arbitrary address could lead to re-entrancy. But do you see a way in how this re-entrancy could exploit something?

TheAustrian1998: Not now, anyway this can be solved using Reentrancy Guard. If you want I can do that change and try to exploit with a reenctrancy later.

nchamo: Let's see what @0xged has to say first, and then we can make a decision together

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -80,9 +80,37 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
...
     uint256 _positionId,
+    address _fromToken,
     uint256 _amount,
     uint32 _newSwaps
-  ) external override {}
+  ) external onlyWithPermission(_positionId, IDCAStrategies.Permission.INCREASE) {
+    Position memory _position = _userPositions[_positionId];
+    IDCAStrategies.ShareOfToken[] memory _tokens = _getTokenShares(_position.strategyId, _position.strategyVersion);
+
+    if (_amount != 0) {
```
nchamo, line 90: Nit: I think this is just a little bit clearer to read
```suggestion
    if (_amount > 0) {
```

nchamo (approved):
Great work!

PR #162, opened 2022-09-08
feat: preparation-for-sync-to-latest

TheAustrian1998:
- create function ` _checkTokenSharesSanity()`
- tests
- prepare code for next PR

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
@@ -4,6 +4,11 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
+  struct StrategyOwnerAndVersion {
+    address owner;
+    uint16 latestVersion;
+  }
+
   /// @inheritdoc IDCAStrategiesManagementHandler
   mapping(uint80 => address) public strategiesPendingOwners;
   /// @inheritdoc IDCAStrategiesManagementHandler
@@ -126,7 +131,7 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
       // isn't the last token
       if (i < _length - 1) {
         // ascending order ([1, 2, 3])
-        if (_tokens[i].token > _tokens[i + 1].token) revert TokenSharesNotSorted(); // not sorted
+        if (_tokens[i].token >= _tokens[i + 1].token) revert TokenSharesNotSorted(); // not sorted
       }
 
       uint16 _share = _tokens[i].share;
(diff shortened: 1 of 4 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -4,11 +4,6 @@ pragma solidity >=0.8.7 <0.9.0;
 import '../../interfaces/IDCAStrategies.sol';
 
 abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHandler {
-  struct StrategyOwnerAndVersion {
```
nchamo, line 7: As discussed offline, we can avoid this

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -195,6 +180,8 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
     IDCAStrategies.Permission _permission
   ) internal view virtual returns (bool _result) {}
 
+  function _getStrategyOwnerAndVersion(uint80 _strategyId) internal view virtual returns (IDCAStrategies.StrategyOwnerAndVersion memory) {}
```
nchamo, line 183: As discussed offline, we can avoid this

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -127,10 +122,20 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
     if (_length > MAX_TOKEN_SHARES) revert TokenSharesExceedAmount(); // need to have less than the max
 
     uint256 _shares = 0;
-    for (uint256 i = 0; i < _length; i++) {
+    for (uint256 i = 0; i < _length; ) {
+      // isn't the last token
+      if (i < _length - 1) {
+        // ascending order ([1, 2, 3])
+        if (_tokens[i].token > _tokens[i + 1].token) revert TokenSharesNotSorted(); // not sorted
```
nchamo, line 129: So that we can detect and fail with duplicates also (please add a test for that too)
```suggestion
        if (_tokens[i].token >= _tokens[i + 1].token) revert TokenSharesNotSorted(); // not sorted
```

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -339,4 +352,18 @@ contract('DCAStrategiesManagementHandler', () => {
       expect(s.token).to.be.equal(arrayB[i].token);
     });
   }
+
+  function sortTokens(array: IDCAStrategies.ShareOfTokenStruct[]) {
+    for (let i = 0; i < array.length; i++) {
```
nchamo, line 357: I think you can sort by a property here, like [this](https://stackoverflow.com/questions/15593850/sort-array-based-on-object-attribute-javascript)

It would be easier to read/understand

nchamo (commented):
Looking good, just a few minor comments

nchamo (approved):
Looks great!

PR #163, opened 2022-09-09
feat: sync-to-latest

TheAustrian1998:
- add code for sync function. I coded the first while of the entire function (from the start to line 70 in the prototype)
- also, I will left gas optimization and tests for the end

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
@@ -110,7 +110,7 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
     uint16 _total = _getTotalShares();
     uint256 _amountSpent;
     for (uint256 i = 0; i < _position.positions.length; ) {
-      uint256 _toIncrease = _calculateOptimalAmount(i < _position.positions.length - 1, _amount, _tokens[i].share, _total, _amountSpent);
+      uint256 _toIncrease = _calculateOptimalAmount(i == _position.positions.length - 1, _amount, _tokens[i].share, _total, _amountSpent);
 
       _position.hub.increasePosition(_position.positions[i], _toIncrease, _newSwaps);
 
@@ -137,7 +137,7 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
     uint16 _total = _getTotalShares();
     uint256 _amountSpent;
     for (uint256 i = 0; i < _position.positions.length; ) {
-      uint256 _toReduce = _calculateOptimalAmount(i < _position.positions.length - 1, _amount, _tokens[i].share, _total, _amountSpent);
+      uint256 _toReduce = _calculateOptimalAmount(i == _position.positions.length - 1, _amount, _tokens[i].share, _total, _amountSpent);
 
       _position.hub.reducePosition(_position.positions[i], _toReduce, _newSwaps, _recipient);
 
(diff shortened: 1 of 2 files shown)


contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -166,7 +178,19 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
   }
 
   /// @inheritdoc IDCAStrategiesPositionsHandler
-  function syncPositionToLatestStrategyVersion(uint256 _positionId) external onlyWithPermission(_positionId, IDCAStrategies.Permission.SYNC) {}
+  function syncPositionToLatestStrategyVersion(
```
nchamo, line 181: I think we can rename this to something like : syncPositionToNewVersion

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -226,15 +250,98 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
...
+    uint256 _totalAmount,
+    IDCAStrategies.ShareOfToken[] memory _newTokenShares,
+    uint32 _newAmountSwaps,
+    address _recipientSwapped
+  ) internal {
+    uint8 _currentPositionsIndex; // old positions index
+    uint8 _newPositionsIndex; // new positions index
+    uint256 _totalRemaining; // cash. The amount of money ready to use. (Used to know if need to send whats left or request whats missing)
+    uint256 _amountSpent;
+    Task[] memory _tasks = new Task[](_newTokenShares.length); // an array containing required tasks
```
nchamo, line 275: The length here should be the sum of the amount of existing positions, and the amount of new token shares

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -226,15 +250,98 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
...
+        _currentPositionsIndex++;
+      } else {
+        // then just create a new position
+        _tasks[_newPositionsIndex] = Task({action: Action.DEPOSIT, positionId: 0, amount: _correspondingToPosition});
+        _newPositionsIndex++;
+      }
+    }
+  }
+
+  function _syncReduceBlock(
```
nchamo, line 322: I'm curious, why the extra function?

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -226,15 +250,98 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
...
+    uint256 _currentPositionId,
+    uint256 _newPositionsIndex
+  ) internal returns (Task[] memory) {
+    _hub.reducePosition(_currentPositionId, _amount, _newAmountSwaps, address(this));
+    _tasks[_newPositionsIndex] = Task({action: Action.REDUCE, positionId: 0, amount: 0});
+
+    return _tasks;
+  }
+
+  function _syncIncreaseBlock(
```
nchamo, line 336: Same here

nchamo (commented):
Overall looks good, just left some comments

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
TheAustrian1998: `Stack too deep` lol

TheAustrian1998: But wait, I think I don't need that. Will try something

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
TheAustrian1998: I remember in the last meeting we discussed that the task array will be the same length as the new shares array

nchamo: Oh, you are right, because we won't be storing "terminate" actions. My bad

nchamo (approved):
Great work!

PR #165, opened 2022-09-20
refactor: delete legacy functions

0xsambugs:
We are now deleting `legacySwapForCaller`, `legacySwapWithDexes` and `legacySwapWithDexesForMean`. We decided that if we want to interact with the legacy Hubs, we will use the old DCA Swapper

### DIFF
--- a/test/integration/DCAHubCompanion/position-migrator.spec.ts
+++ b/test/integration/DCAHubCompanion/position-migrator.spec.ts
@@ -4,7 +4,7 @@ import { TransactionResponse } from '@ethersproject/providers';
 import { constants, wallet } from '@test-utils';
 import { given, then, when } from '@test-utils/bdd';
 import evm, { snapshot } from '@test-utils/evm';
-import { IERC20, DCAHubCompanion, DCAHubSwapper } from '@typechained';
+import { IERC20, DCAHubCompanion, DCAHubSwapper, LegacyDCASwapper, LegacyDCASwapper__factory } from '@typechained';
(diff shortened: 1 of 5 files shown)


test/integration/DCAHubCompanion/position-migrator.spec.ts
```diff
@@ -89,10 +88,10 @@ describe('Position Migration', () => {
     // Send tokens from whales, to our users
     await distributeTokensToUsers();
 
-    // Approve swapper
-    await DCAHubSwapper.connect(msig).grantRole(await DCAHubSwapper.SWAP_EXECUTION_ROLE(), swapper.address);
-
     chainId = BigNumber.from((await ethers.provider.getNetwork()).chainId);
+
+    const legacyDCAHubSwapperFactory: LegacyDCASwapper__factory = await ethers.getContractFactory('LegacyDCASwapper');
+    legacyDCAHubSwapper = await legacyDCAHubSwapperFactory.deploy();
```
0xged, line 94: Not entirely sure why we are deploying and using `Legacy` here instead of using the interface and the on-chain swapper? Am I missing something? 😬

0xged (approved):
Other than the comment - looks good!

test/integration/DCAHubCompanion/position-migrator.spec.ts
0xsambugs: The on-chain swapper has a hardcoded hub. We want to execute swaps for all previous versions, so we needed to deploy a new contract for that. But since we are only using it here, it's under `mocks`

PR #166, opened 2022-09-20
feat: add intermediate tokens

0xsambugs:
We realized that in complex swaps (like yield(WETH) => WETH => USDC), we might have intermediate tokens. But we don't have a way to check these tokens for unspent balance. So we are adding this


0xged (approved):
The amount of value being added by our pre-deploy tests. Great work!

PR #168, opened 2022-09-20
feat: sync-to-latest-tests

TheAustrian1998:
- add tests for sync function
- fix errors

### DIFF
--- a/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
+++ b/contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
@@ -198,7 +198,27 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
     Position memory _position = _userPositions[_positionId];
     IDCAStrategies.ShareOfToken[] memory _newTokenShares = _getTokenShares(_position.strategyId, _newVersion);
 
+    // for (uint256 e = 0; e < _position.positions.length; e++) {
+    //   console.log('positions i:', e, 'val: ', _position.positions[e]);
+    // }
+
+    for (uint256 x = 0; x < _newTokenShares.length; x++) {
+      console.log('shares: i: ', x, 'share: ', _newTokenShares[x].token);
+      console.log('||', _newTokenShares[x].share);
+    }
+
+    console.log('about to enter in _syncBlock', _totalAmount, _newAmountSwaps, _recipientSwapped);
+
     (Data memory _data, Task[] memory _tasks) = _syncBlock(_position, _totalAmount, _newTokenShares, _newAmountSwaps, _recipientSwapped);
+
+    console.log('just out of _syncBlock data:', _data.currentPositionsIndex, _data.newPositionsIndex);
+    // console.log('||', _data.totalRemaining, _data.amountSpent);
+    // console.log('user position tokens: ', address(_data.positionMetadata.from), address(_data.positionMetadata.to));
+    // console.log('|| numbers: ', _data.positionMetadata.swapInterval, _data.positionMetadata.swapsExecuted, _data.positionMetadata.swapped);
+
+    // for (uint256 s = 0; s < _tasks.length; s++) {
+    //   console.log('tasks first', uint256(_tasks[s].action), _tasks[s].amount, _tasks[s].positionId);
+    // }
 
     // If get to this place, then we just need to terminate existing positions
     while (_data.currentPositionsIndex < _position.positions.length) {
@@ -207,6 +227,7 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
       _data.currentPositionsIndex++;
     }
 
+    // If get to this place, then we just need to deposit
     while (_data.newPositionsIndex < _newTokenShares.length) {
       IDCAStrategies.ShareOfToken memory _newTokenShare = _newTokenShares[_data.newPositionsIndex];
       uint256 _correspondingToPosition = _calculateOptimalAmount(
@@ -216,7 +237,7 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
         _getTotalShares(),
         _data.amountSpent
       );
-      _tasks[_data.newPositionsIndex] = Task({action: Action.DEPOSIT, amount: _correspondingToPosition, positionId: _positionId});
+      _tasks[_data.newPositionsIndex] = Task({action: Action.DEPOSIT, amount: _correspondingToPosition, positionId: 0});
       _data.amountSpent += _correspondingToPosition;
       _data.newPositionsIndex++;
     }
@@ -228,9 +249,14 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
       IERC20(_data.positionMetadata.from).safeTransfer(_recipientUnswapped, _data.totalRemaining - _totalAmount);
     }
 
+    for (uint256 s = 0; s < _tasks.length; s++) {
+      console.log('tasks second', uint256(_tasks[s].action), _tasks[s].amount, _tasks[s].positionId);
+    }
+
     // perform deposit and increase
     for (uint256 i = 0; i < _tasks.length; ) {
       Task memory _task = _tasks[i];
+      uint256 _auxPositionId = _positionId;
 
       if (_task.action == Action.INCREASE) {
         _position.hub.increasePosition(_task.positionId, _task.amount, _newAmountSwaps);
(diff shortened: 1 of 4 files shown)


test/unit/DCAStrategies/dca-strategies-positions-handler.spec.ts
```diff
@@ -445,4 +456,211 @@ contract('DCAStrategiesPositionsHandler', () => {
...
+
+          newPositions = (await DCAStrategiesPositionsHandlerMock.userPosition(1)).positions;
+        });
+        then('transfer() is called correctly', async () => {
+          expect(tokenF.transfer).to.have.been.calledWith(user.address, delta);
+        });
+        then('increase, reduce, deposit or terminate is called correctly', async () => {
+          checkCallsToHub();
+          expect(hub.reducePosition).to.have.been.calledTwice;
+          expect(hub.increasePosition).to.have.callCount(0);
```
nchamo, line 544:
```suggestion
          expect(hub.increasePosition).to.not.have.been.called
```

test/unit/DCAStrategies/dca-strategies-positions-handler.spec.ts
```diff
@@ -445,4 +456,211 @@ contract('DCAStrategiesPositionsHandler', () => {
...
+          tx = await callSync(totalAmount.sub(delta), newAmountOfSwaps);
+
+          newPositions = (await DCAStrategiesPositionsHandlerMock.userPosition(1)).positions;
+        });
+        then('transfer() is called correctly', async () => {
+          expect(tokenF.transfer).to.have.been.calledWith(user.address, delta);
+        });
+        then('increase, reduce, deposit or terminate is called correctly', async () => {
+          checkCallsToHub();
+          expect(hub.reducePosition).to.have.been.calledTwice;
```
nchamo, line 543: Could we check that each reduce was called with the correct amount? I know it's a pain, but I think it's very important to understand that we are doing everything correctly

test/unit/DCAStrategies/dca-strategies-positions-handler.spec.ts
```diff
@@ -445,4 +456,211 @@ contract('DCAStrategiesPositionsHandler', () => {
...
+        given(async () => {
+          tx = await callSync(totalAmount.sub(delta), newAmountOfSwaps);
+
+          newPositions = (await DCAStrategiesPositionsHandlerMock.userPosition(1)).positions;
+        });
+        then('transfer() is called correctly', async () => {
+          expect(tokenF.transfer).to.have.been.calledWith(user.address, delta);
+        });
+        then('increase, reduce, deposit or terminate is called correctly', async () => {
+          checkCallsToHub();
```
nchamo, line 542: Could we check that `deposit` and `terminate` were called with the correct parameters?

test/unit/DCAStrategies/dca-strategies-positions-handler.spec.ts
```diff
@@ -445,4 +456,211 @@ contract('DCAStrategiesPositionsHandler', () => {
...
+        ).to.be.revertedWith('NoPermissions()');
+      });
+    });
+    when('syncPositionToNewVersion is called', () => {
+      given(async () => {
+        const positions = [1, 2, 3];
+        const OLD_SHARE_A = { token: tokenA.address, share: BigNumber.from(33e2) }; // reduce
+        const OLD_SHARE_B = { token: tokenB.address, share: BigNumber.from(33e2) }; // increase
+        const OLD_SHARE_C = { token: tokenC.address, share: BigNumber.from(34e2) }; // terminate
+        oldShares = sortTokens([OLD_SHARE_A, OLD_SHARE_B, OLD_SHARE_C]);
```
nchamo, line 486: Could we add a token whose shares doesn't change? It would be interesting to understand that it works correctly

nchamo (commented):
Amazing work @TheAustrian1998 !

Left some comments

test/unit/DCAStrategies/dca-strategies-management-handler.spec.ts
```diff
@@ -368,12 +368,19 @@ contract('DCAStrategiesManagementHandler', () => {
...
-      if (parseInt(a.token, 16) > parseInt(b.token, 16)) return 1;
+      if (hexToNumber(a.token) < hexToNumber(b.token)) return -1;
+      if (hexToNumber(a.token) > hexToNumber(b.token)) return 1;
+      if (hexToNumber(a.token) == hexToNumber(b.token)) console.error('found duplicate when sorting');
       return 0;
     }
 
-    return array.sort(compare);
+    let f = array.sort(compare);
+    console.log('array f', f);
```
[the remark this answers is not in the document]
TheAustrian1998: My bad

test/unit/DCAStrategies/dca-strategies-positions-handler.spec.ts
TheAustrian1998: good catch, I wasn't accounting for a edge-case where nothing changes

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -371,6 +372,9 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
             positionId: _currentPositionId,
             amount: _correspondingToPosition - _userPosition.remaining
           });
+        } else {
+          // do nothing
+          _tasks[_data.newPositionsIndex] = Task({action: Action.NOTHING, positionId: _currentPositionId, amount: 0});
```
nchamo, line 377: Is this necessary? Since `NOTHING` is the first value in the enum, it will be the default value, so I think it might not be necessary to assign it

test/unit/DCAStrategies/dca-strategies-positions-handler.spec.ts
```diff
@@ -445,4 +458,256 @@ contract('DCAStrategiesPositionsHandler', () => {
...
+      then('tx reverts with message', async () => {
+        await expect(
+          DCAStrategiesPositionsHandlerMock.syncPositionToNewVersion(1, 3, user.address, user.address, amountOfSwaps, newAmountOfSwaps)
+        ).to.be.revertedWith('NoPermissions()');
+      });
+    });
+    when('syncPositionToNewVersion is called', () => {
+      given(async () => {
+        const positions = [1, 2, 3];
+        const OLD_SHARE_A = { token: tokenA.address, share: BigNumber.from(33e2) }; // nothing
```
nchamo, line 486: But now we are not testing reduce 😅 

Could we try to test all situations? Maybe adding an extra token?

nchamo (commented):
Minor comments

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
TheAustrian1998: I think it is. I need it to save the positionId to be assigned in the last for loop. I can reuse existing enum values if you want. (In line 377 when writing the task I could use the REDUCE value or any other. It doesn't matter because is not read).

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -241,19 +247,17 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
...
       }
 
+      _newPositions[i] = _task.positionId;
+
       unchecked {
         i++;
       }
     }
 
+    _userPositions[_positionId].positions = _newPositions;
```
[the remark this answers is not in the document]
TheAustrian1998: I noticed some strange behaviors that caused an ID to be overwritten (i.e. a position that was reused was overridden by a new position just deposited). Also, it made the code much more simpler 😅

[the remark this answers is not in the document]
nchamo: I agree that the code is **much much** simpler, but I think it will end up being more expensive. Because in the case of increase/reduce/nothing, you are re-writing the same value that was already there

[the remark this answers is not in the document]
TheAustrian1998: What do you suggest?

contracts/DCAStrategies/DCAStrategies/DCAStrategiesPositionsHandler.sol
```diff
@@ -229,7 +229,7 @@ abstract contract DCAStrategiesPositionsHandler is IDCAStrategiesPositionsHandle
       IERC20(_data.positionMetadata.from).safeTransfer(_recipientUnswapped, _data.totalRemaining - _totalAmount);
     }
 
-    uint256[] memory _newPositions = new uint256[](_tasks.length);
+    uint256 _auxPositionId = _positionId;
```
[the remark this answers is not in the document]
TheAustrian1998: Remember that weird `Stack too deep` error? That's the way of fixing it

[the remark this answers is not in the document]
nchamo: OMG Solidity, why you like this?

nchamo (approved):
Amazing work @TheAustrian1998 !

This was a really long and hard PR, but you make it work 💪💪💪 

Thank you so much for this!

PR #169, opened 2022-09-22
perf: some minor optimizations

0xsambugs:
We are now using unchecked in for loops, and we avoiding unnecessary array reads to make some small gas savings


contracts/DCAFeeManager/DCAFeeManager.sol
```diff
@@ -121,11 +133,11 @@ contract DCAFeeManager is RunSwap, AccessControl, Multicall, IDCAFeeManager {
   /// @inheritdoc IDCAFeeManager
   function availableBalances(IDCAHub _hub, address[] calldata _tokens) external view returns (AvailableBalance[] memory _balances) {
     _balances = new AvailableBalance[](_tokens.length);
-    for (uint256 i; i < _tokens.length; i++) {
+    for (uint256 i = 0; i < _tokens.length; i++) {
```
nchamo, line 136: Why not use unchecked in these?

0xsambugs: Because `availableBalances` is meant to be used off-chain

0xged (approved):

nchamo (approved):

PR #170, opened 2022-09-26
feat: size-optimizations

TheAustrian1998:


contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
```diff
@@ -126,7 +132,7 @@ abstract contract DCAStrategiesManagementHandler is IDCAStrategiesManagementHand
     if (_length <= 1) revert InvalidLength(); // need to have more than one item
     if (_length > MAX_TOKEN_SHARES) revert TokenSharesExceedAmount(); // need to have less than the max
 
-    uint256 _shares = 0;
```
nchamo, line 129: Did you see a size reduce with this change? I think. that the compile just ignores the `= 0`

nchamo (approved):
Great work! How much size did we get back?

contracts/DCAStrategies/DCAStrategies/DCAStrategiesManagementHandler.sol
TheAustrian1998: I see no changes, but I removed it because it wasn't necessary

PR #178, opened 2022-11-18
feat: add set swapper to keep3r job

0xsambugs:
We are now adding `setSwapper` to the keep3r job

### DIFF
--- a/test/unit/DCAKeep3rJob/dca-keep3r-job.spec.ts
+++ b/test/unit/DCAKeep3rJob/dca-keep3r-job.spec.ts
@@ -1,10 +1,11 @@
 import chai, { expect } from 'chai';
 import { ethers } from 'hardhat';
 import { behaviours, constants } from '@test-utils';
-import { contract, then, when } from '@test-utils/bdd';
+import { contract, given, then, when } from '@test-utils/bdd';
 import { snapshot } from '@test-utils/evm';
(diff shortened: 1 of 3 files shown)


test/unit/DCAKeep3rJob/dca-keep3r-job.spec.ts
```diff
@@ -97,4 +98,41 @@ contract('DCAKeep3rJob', () => {
...
+        tx = await DCAKeep3rJob.connect(superAdmin).setSwapper(NEW_SWAPPER);
+      });
+      then('it is set correctly', async () => {
+        const { swapper } = await DCAKeep3rJob.swapperAndNonce();
+        expect(swapper).to.equal(NEW_SWAPPER);
+      });
+      then('nonce is not modified', async () => {
+        const { nonce } = await DCAKeep3rJob.swapperAndNonce();
+        expect(nonce).to.equal(constants.ZERO);
+      });
```
0xged, line 125: Should we test this with `nonce != 0`? Although... if nonce is changed when swapper gets changed... It's an issue with the EVM, not with our code lmao.

0xged (approved):
Great work Sam! LFG!

test/unit/DCAKeep3rJob/dca-keep3r-job.spec.ts
0xsambugs: Not sure I understand the suggestion. It was zero before the swapper was set, want to check it's still zero after it's set

PR #179, opened 2022-11-21
feat: add keep3r job work

0xsambugs:
We are now implementing the job's `work` function

### DIFF
--- a/test/unit/DCAKeep3rJob/dca-keep3r-job.spec.ts
+++ b/test/unit/DCAKeep3rJob/dca-keep3r-job.spec.ts
@@ -1,13 +1,15 @@
 import chai, { expect } from 'chai';
 import { ethers } from 'hardhat';
-import { behaviours, constants } from '@test-utils';
+import { behaviours, constants, wallet } from '@test-utils';
 import { contract, given, then, when } from '@test-utils/bdd';
 import { snapshot } from '@test-utils/evm';
-import { DCAKeep3rJob, DCAKeep3rJob__factory, IDCAHubSwapper, IKeep3r } from '@typechained';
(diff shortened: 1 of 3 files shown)


test/unit/DCAKeep3rJob/dca-keep3r-job.spec.ts
```diff
@@ -135,4 +143,152 @@ contract('DCAKeep3rJob', () => {
...
+      const { domain, types, value } = buildWorkData(options);
+      const signature = await options.signer._signTypedData(domain, types, value);
+      return fromRpcSig(signature);
+    }
+
+    function buildWorkData(options: OperationData) {
+      return {
+        primaryType: 'Work',
+        types: { Work },
+        domain: { name: 'Mean Finance - DCA Keep3r Job', version: '1', chainId: options.chainId, verifyingContract: DCAKeep3rJob.address },
```
0xged, line 270: Just realized while reading this... Seems sensible to make the `domain` much shorter. `MF - DCAKJ` maybe? Since every byte being hashed on-chain to check signer will make it more gas-expensive. WDYT?

0xged (approved):
Amazing work Sam! Love the parameterized failing tests 👏🏽

test/unit/DCAKeep3rJob/dca-keep3r-job.spec.ts
0xsambugs: That string is hashed only during the deployment. It's not hashed on each verify, so we would only be saving a little gas on the deployment. Since that's the case, I'd prefer to make it more readable, if it's ok with you of course

PR #184, opened 2023-02-12
feat: add third party dca swapper

0xsambugs:
We are now introducing a new DCA Swapper. This Swapper is meant to only support third-party swaps, not swaps for caller
This means that no-one should be approving tokens for this contract. Since that's the case, and the contract is not meant to hold any tokens, we can call any arbitrary data

Therefore, we can avoid using the TransformerRegistry and interact with 4626s directly 🚀 

Also, in the existing swapper, the workflow is DCASwapper => DCAHub => DCASwapper

Now, we only do DCAHub => DCASwapper, which should be much cheaper in terms of gas 🤑

### DIFF
--- a/contracts/DCAHubSwapper/ThirdPartyDCAHubSwapper.sol
+++ b/contracts/DCAHubSwapper/ThirdPartyDCAHubSwapper.sol
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity >=0.8.7 <0.9.0;
+
+import '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
+import '@mean-finance/dca-v2-core/contracts/interfaces/IDCAHubSwapCallee.sol';
+
+contract ThirdPartyDCAHubSwapper is IDCAHubSwapCallee {
+  /// @notice A target we want to give allowance to
+  struct Allowance {
+    IERC20 token;
+    address spender;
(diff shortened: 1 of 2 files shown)


contracts/DCAHubSwapper/ThirdPartyDCAHubSwapper.sol
```diff
@@ -0,0 +1,129 @@
...
+
+  // solhint-disable-next-line func-name-mixedcase
+  function DCAHubSwapCall(
+    address,
+    IDCAHub.TokenInSwap[] calldata _tokens,
+    uint256[] calldata,
+    bytes calldata _data
+  ) external {
+    SwapWithDexesCallbackData memory _callbackData = abi.decode(_data, (SwapWithDexesCallbackData));
+    if (block.timestamp > _callbackData.deadline) revert TransactionTooOld();
```
0xged, line 52: Do you think we really need this? Most of the trades are going to be routed through places that already have a deadline check.

0xged (approved):
No one should ever ever approve this DCAHubSwapper. Having said that, it looks good! 🚀

contracts/DCAHubSwapper/ThirdPartyDCAHubSwapper.sol
0xsambugs: Only a few due, and by checking it ourselves we can fail fast (and cheap)

PR #187, opened 2023-02-12
refactor: remove dexes code from old swapper

0xsambugs:
We are now removing all the code to swap with dexes from the old swapper. We will only leave the code to swap for caller


0xged (approved):
The sweet smell of taking code out from a smart contract! Looks good, much more simplified. Great idea on splitting this two.

PR #188, opened 2023-02-12
refactor: rename old swapper

0xsambugs:
We are renaming the old swapper so that it's clear that it will only support "swap for caller" only

### DIFF
--- a/deploy/002_caller_only_swapper.ts
+++ b/deploy/002_caller_only_swapper.ts
@@ -1,6 +1,6 @@
 import { HardhatRuntimeEnvironment } from 'hardhat/types';
 import { DeployFunction } from '@0xged/hardhat-deploy/types';
-import { bytecode } from '../artifacts/contracts/DCAHubSwapper/DCAHubSwapper.sol/DCAHubSwapper.json';
+import { bytecode } from '../artifacts/contracts/DCAHubSwapper/CallerOnlyDCAHubSwapper.sol/CallerOnlyDCAHubSwapper.json';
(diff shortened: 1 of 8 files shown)


deploy/002_caller_only_swapper.ts
```diff
@@ -10,9 +10,9 @@ const deployFunction: DeployFunction = async function (hre: HardhatRuntimeEnviro
 
   await deployThroughDeterministicFactory({
     deployer,
-    name: 'DCAHubSwapper',
-    salt: 'MF-DCAV2-DCAHubSwapper-V1',
-    contract: 'contracts/DCAHubSwapper/DCAHubSwapper.sol:DCAHubSwapper',
+    name: 'DCAHubSwapper', // We will use the old name to avoid re-deploying
```
0xged, line 13: Why would we not re-deploy this one? Idea is to reuse the old one for swapping for caller?

0xsambugs: Yeah, that's the plan. Now that you say it, we should use the same salt so that the contracts have the same address. Since the code is the same for 'swap for caller', then we can re-use the currently deployed versions
What do you think?

0xged: Oh I see, so basically maintaining the address, interface and compatibility, but just stripping out everything from the contract.  I mean, I do understand the "why" - but for the sake of consistency I'd change the salt, the name and re-deploy (everything is cheap af, except for Ethereum that it will not be *that* expensive). But that is just who I am and my small OCD with deployment addresses, and consistency all around.

0xsambugs: Ok, will do! 👍

PR #189, opened 2023-03-10
feat: add execute swap

0xsambugs:
There are some cases where the oracles differ from what the markets can offer, so a swap can't be executed. But it could happen that even if the amounts being swap are really big, the difference between oracle and market is only a few dollars. In that case, it would be nice if someone could just pay for the difference.
The idea here is that instead of calling the hub directly, someone could call the swapper with some native token, so that when the swapper gets called, they can use that native token balance as part of the swap, and cover the difference

The normal workflow is:
`Caller => Hub => Swapper => Hub`

Here, it would be:
`Caller => Swapper => Hub => Swapper => Hub`

It would be more expensive, but again, it would be optional

_Note: this change required us to include a new property `value` in `SwapExecution`. It would make normal swaps (that start on the Hub) a little more expensive. It's hard to understand how much, since we can't do gas tests on calls that start on the hub (I hate you hardhat), but we would only be adding 1 uint full of zeros to the calldata_

### DIFF
--- a/contracts/DCAHubSwapper/ThirdPartyDCAHubSwapper.sol
+++ b/contracts/DCAHubSwapper/ThirdPartyDCAHubSwapper.sol
@@ -15,6 +15,7 @@ contract ThirdPartyDCAHubSwapper is IDCAHubSwapCallee {
   /// @notice The data necessary for a swap to be executed
   struct SwapExecution {
     address swapper;
+    uint256 value;
     bytes swapData;
   }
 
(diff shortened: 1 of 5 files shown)


test/integration/DCAHubSwapper/swap-with-dex-native.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+        expect(await performedSwaps()).to.equal(initialPerformedSwaps + 1);
+      });
+      then('hub balance is correct', async () => {
+        const hubWETHBalance = await WETH.balanceOf(DCAHub.address);
+        const hubUSDCBalance = await USDC.balanceOf(DCAHub.address);
+        expect(hubWETHBalance).to.equal(initialHubWETHBalance.sub(reward));
+        expect(hubUSDCBalance).to.equal(initialHubUSDCBalance.add(receivedFromAgg));
+      });
+      then('all reward is sent to leftover recipient', async () => {
+        const recipientWETHBalance = await WETH.balanceOf(recipient.address);
```
0xsambugs, line 125: See that all WETH is sent to the recipient, since we spent ETH for the swap

0xged: Ohhhhh. It was a really delicate change. Thanks for pointing out, I would have missed it.

contracts/DCAHubSwapper/ThirdPartyDCAHubSwapper.sol
```diff
@@ -56,6 +57,26 @@ contract ThirdPartyDCAHubSwapper is IDCAHubSwapCallee {
...
+  function executeSwap(
+    IDCAHub _hub,
+    address[] calldata _tokens,
+    IDCAHub.PairIndexes[] calldata _pairsToSwap,
+    uint256[] calldata _borrow,
+    bytes calldata _callbackData,
+    bytes calldata _oracleData
+  ) external payable {
+    _hub.swap(_tokens, _pairsToSwap, address(this), address(this), _borrow, _callbackData, _oracleData);
+  }
```
0xged, line 78: This contract should always be approved exact amount by callers, right? If not, then... This function is dangerous since `_hub` can be any contract.

0xsambugs: There is no approval needed. This function is meant the be used for market swaps + native balance, which doesn't need approval

0xged: Oh my god, my brain f*cked up. I was thinking of the `CallerOnly`.

0xged (approved):
Looks good! Other approach could be sending ERC20 tokens + sending the trade that utilizes that also. But it's good to go for me! 🚀

0xsambugs:
But that would require approvals or more than one tx. That's why I went with this approach

PR #193, opened 2023-04-09
feat: fee manager withdraw balance

0xsambugs:
We are now adding a way to withdraw all the balance from the Fee Manager. Before this change, we would need to specify a specific amount, which was kind of a buffer if we executed a swap and couldn't predict how much we would get in response


.github/workflows/tests.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         run: yarn test:integration
         env:
           TS_NODE_SKIP_IGNORE: true
-          ETH_NODE_URI_ETHEREUM: https://rpc.ankr.com/eth
-          ETH_NODE_URI_OPTIMISM: https://rpc.ankr.com/optimism
+          ETH_NODE_URI_ETHEREUM: https://eth-mainnet.g.alchemy.com/v2/${{ secrets.ALCHEMY_API_KEY }}
```
0xsambugs, line 74: Tests weren't working with Ankr

test/integration/DCAHubSwapper/single-pair-swap-with-dex.spec.ts
```diff
@@ -1,262 +0,0 @@
-import { expect } from 'chai';
```
0xsambugs, line 1: Needed to reduce the amount of tests so we didn't get rate-limited, and since we already have the multi-pair-swap test, decided to delete this one

0xged (approved):
Looks pretty straight forward! Thanks for fixing the tests!

PR #198, opened 2023-07-28
refactor: remove swapper registry from fee manager

0xsambugs:
We are now removing the Swapper Registry from the Fee Manager. Since all calls are permissioned, we don't need the registry. We are:
- Removing `runSwap`
- Chainging `takeManyRunSwapsAndTransferMany` to `runSwapsAndTransferMany`
  - This version does not take tokens from the caller
  - And does not use the swapper registry when checking for swappers or allowance targets


0xged (approved):
Looks gucci! LFG!

PR #200, opened 2023-07-31
feat: add simulation adapter

0xsambugs:
We are now adding the new Simulation Adapter to the companion, so that we can simulate multiple quotes before executing them


0xged (approved):
This is just marvelous! 🚀

PR #205, opened 2024-02-05
refactor: remove min amount out from base companion

0xsambugs:
Now that we've introduced [this change](https://github.com/Mean-Finance/permit2-adapter/pull/33), we can call `sellOrderSwap` and `buyOrderSwap`, that already perform this check


0xKoaj (approved):
Nice work!

PR #207, opened 2024-03-08
refactor: simplify fee manager

0xsambugs:
We'll remove code we are not using, to lower contract size and deployment costs


0xKoaj (approved):
Wow, nice pruning!

PR #208, opened 2024-03-18
feat: add simulation adapter to fee manager

0xsambugs:
We are now adding simulation capabilities with the fee manager


0xKoaj (approved):
Looks good!

